PR #1, opened 2023-06-08
Add README

9999years:


evanrelf:
@9999years I don't think the reasons you listed for writing this in Rust are very compelling.

- Shipping static binaries is useful for bootstrapping (necessary for `bootstrap-mercury`) but we're gonna just include this in our Nix shell, so that doesn't really matter.
- `watchexec` is cool, but [`streamly-fsnotify`](https://hackage.haskell.org/package/streamly-fsnotify) is also excellent, and I think would meet all your file watching needs.
- It makes more sense for a Haskell-specific tool to be written in Haskell, considering the community you're serving would be most comfortable using and contributing to a project in that language.

I have some experience playing around with my own "next generation" `ghcid`, and Haskell and its libraries were not a limiting factor whatsoever. `streamly`, `streamly-fsnotify`, `brick`, `lsp`, etc. were all mature libraries that met my needs.

That said, I think if you strongly prefer writing this in Rust, and you enjoy it and are more productive, that's reason enough 🙂

9999years:
@evanrelf `watchexec` is really the compelling factor here, and I think it'll cut down on implementation difficulty significantly:

- `watchexec` supports [runtime reconfiguration](https://docs.rs/watchexec/latest/watchexec/struct.Watchexec.html#method.reconfigure); Haskell's `fsnotify` doesn't. (This is one of the reasons `ghcid` performs so poorly around renames, new directories, etc.)
- `watchexec` has an event-loop built in, with support for [sleeping, clearing the screen, etc.](https://docs.rs/watchexec/latest/watchexec/action/enum.Outcome.html) `fsnotify` doesn't, so I'd need to reimplement all this functionality in Haskell.
- `watchexec` has support for [running commands on file changes](https://docs.rs/watchexec/latest/watchexec/config/struct.RuntimeConfig.html#method.command), which will make it easy to implement test-on-change or `hpack`-on-change.
- `watchexec` has a really slick system of [tagged events](https://docs.rs/watchexec-events/1.0.0/watchexec_events/enum.Tag.html) which delivers a lot more granular information than what Haskell's `fsnotify` [is capable of](https://hackage.haskell.org/package/fsnotify-0.4.1.0/docs/System-FSNotify.html#t:Event). This also provides a unified interface for handling file changes, keyboard events (like EOF), and signals from subprocesses.
- `watchexec` is significantly more mature than the Haskell equivalents I've seen. It's [widely packaged](https://repology.org/project/watchexec/versions), there's [dozens of contributors](https://github.com/watchexec/watchexec/graphs/contributors) actively working on the project (meanwhile `fsnotify` hasn't seen a commit since last year), and it's battle-tested in the `watchexec` tool itself, but also in [`cargo-watch`](https://github.com/watchexec/cargo-watch) and [Tectonic](https://github.com/tectonic-typesetting/tectonic).

Are there some killer Haskell libraries for this stuff that I could use instead? I looked around but couldn't find much. I've also gotten the impression that the Rust libraries for a lot of command-line stuff are much more polished than their Haskell equivalents (look at how much care is taken by `watchexec` just to [clear the terminal screen](https://github.com/watchexec/clearscreen/blob/main/src/lib.rs)). I'd love to be proven wrong here.

> It makes more sense for a Haskell-specific tool to be written in Haskell.

I do agree with this. [Haskell bindings and Hackage distribution are goals for this project](https://linear.app/mercury/issue/DUX-1218/haskell-bindings), but of course that won't change the implementation language.

> You'd be missing out on Haskell-exclusive libraries like `ghc-lib` / `ghc-lib-parser` that may be useful for supplementing what you get from `ghci`.

I don't _think_ we'll need these. Even `ghcid` just parses the `ghci` output. Using `ghc-lib[-parser]` also makes libraries brittle, difficult to upgrade, and precludes supporting multiple versions of GHC with one codebase or one executable entirely.

evanrelf:
Hey wiggles, thanks for such a thorough explanation. I hope I'm not sounding antagonistic or anything; I don't mean to suggest writing this in Rust is a bad idea. It sounds like you've put a lot of thought into this. I'm looking forward to what gets built 🙂 

---

> `watchexec` is really the compelling factor here

Dang `watchexec` seems pretty awesome. I didn't know it did all of that.

To clarify, I'm not suggesting `streamly-fsnotify` would do all of that. I was thinking of `streamly-fsnotify` as a piece of a larger system. You'd handle drawing to the screen, keyboard events, etc. separately from filesystem events, using `brick` as your declarative TUI library.

Maybe you can get away with just using `watchexec` for your user interface, but I suspect you'll want something like `tui` or `cursive` later on for more sophisticated UI stuff\*.



It's cool that `watchexec` a batteries-included framework, but if you outgrow it, you're gonna end up with a similar design to what I described in Haskell anyways.

> Using `ghc-lib[-parser]` also makes libraries brittle, difficult to upgrade, and precludes supporting multiple versions of GHC

Correct me if I'm wrong, but those aren't problems with `ghc-lib{,-parser}`, those are problems with `ghc`. I specifically mentioned `ghc-lib{,-parser}` because they work across multiple versions of GHC, and you can use newer GHC APIs with older versions of GHC.

---

\* I find `ghcid`'s "print `ghci` output verbatim" disappointing; there's so much you could do with the UI if you digested `ghci`'s output more, like sorting diagnostics by severity (like [`cargo-limit`](https://github.com/alopatindev/cargo-limit)), interactively collapsing diagnostics with a [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details)-style UI, showing output from programs like `hlint` alongside GHC's output... the list goes on. There's so many cool things you could do beyond regurgitating `ghci` output.

9999years:
@evanrelf 

> I hope I'm not sounding antagonistic or anything; I don't mean to suggest writing this in Rust is a bad idea.

None taken. It's important to hammer out considerations like this, and I _am_ a bit hesitant to write it in Rust for the same reasons, but I also think Rust will be the fastest implementation language (in terms of time-to-MVP), and I don't want to spend much more time on this than I have to.

> To clarify, I'm not suggesting streamly-fsnotify would do all of that. I was thinking of streamly-fsnotify as a piece of a larger system. You'd handle drawing to the screen, keyboard events, etc. separately from filesystem events, using brick as your declarative TUI library.

Yeah, understood. I don't think I'll be using `watchexec` for drawing to the screen either. (Although the MVP of this will probably be minimal enough that I'm not doing any terminal drawing other than passing `ghci` output through.)

> You can use newer GHC APIs with older versions of GHC [with the `ghc-lib-parser` libraries].

Yeah, I had this wrong.

> I find ghcid's "print ghci output verbatim" disappointing; there's so much you could do with the UI if you digested ghci's output more, like sorting diagnostics by severity (like [cargo-limit](https://github.com/alopatindev/cargo-limit)), interactively collapsing diagnostics with a [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details)-style UI, showing output from programs like hlint alongside GHC's output... the list goes on. There's so many cool things you could do beyond regurgitating ghci output.

Yeah, these are good ideas for the future.

PR #8, opened 2023-08-08
Log errors in async tasks

9999years:
This will actually log errors using `tracing::error!`. No more silent errors!

The `backoff` crate does claim to support futures, but I couldn't get it to work with my functions. Something about calling an `&mut self` function, I think? Anyways, I can at least use their exponential backoff implementation, which is quite nice: it supports duration caps on both individual retry attempts and the total elapsed time, adjustable jitter, and fractional exponents.

I'm not in love with the boilerplate duplicated between the three main tasks, but it may have to do.

If I try to copy the [`backoff` async example](https://github.com/ihrwein/backoff/blob/master/examples/async.rs) like so:

```rust
backoff::future::retry(ExponentialBackoff::default(), || async move {
    match self.run_inner().await {
        Ok(()) => {
            // MPSC channel closed, probably a graceful shutdown?
            tracing::debug!("Channel closed");
            Ok(())
        }
        Err(err) => {
            tracing::error!("{err:?}");
            Err(backoff::Error::transient(err))
        }
    }
})
.await
```

I get this error:

```
error[E0507]: cannot move out of `self`, a captured variable in an `FnMut` closure
  --> src/ghci/stdin.rs:48:66
   |
47 |       pub async fn run(mut self) -> miette::Result<()> {
   |                        -------- captured outer variable
48 |           backoff::future::retry(ExponentialBackoff::default(), || async move {
   |  _______________________________________________________________--_^
   | |                                                               |
   | |                                                               captured by this `FnMut` closure
49 | |             match self.run_inner().await {
   | |                   ----
   | |                   |
   | |                   variable moved due to use in generator
   | |                   move occurs because `self` has type `GhciStdin`, which does not implement the `Copy` trait
50 | |                 Ok(()) => {
51 | |                     // MPSC channel closed, probably a graceful shutdown?
...  |
59 | |             }
60 | |         })
   | |_________^ `self` is moved here
```

If I use an `async` block instead of an `async move` block, I get this error:

```
error: captured variable cannot escape `FnMut` closure body
  --> src/ghci/stdin.rs:48:66
   |
47 |       pub async fn run(mut self) -> miette::Result<()> {
   |                        -------- variable defined here
48 |           backoff::future::retry(ExponentialBackoff::default(), || async {
   |  ________________________________________________________________-_^
   | |                                                                |
   | |                                                                inferred to be a `FnMut` closure
49 | |             match self.run_inner().await {
   | |                   ---- variable captured here
50 | |                 Ok(()) => {
51 | |                     // MPSC channel closed, probably a graceful shutdown?
...  |
59 | |             }
60 | |         })
   | |_________^ returns an `async` block that contains a reference to a captured variable, which then escapes the closure body
   |
   = note: `FnMut` closures only have access to their captured variables while they are executing...
   = note: ...therefore, they cannot allow references to captured variables to escape
```

### DIFF
--- a/src/ghci/mod.rs
+++ b/src/ghci/mod.rs
@@ -6,6 +6,7 @@ use std::process::Stdio;
 use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
 
+use aho_corasick::AhoCorasick;
 use camino::Utf8PathBuf;
 use itertools::Itertools;
 use miette::IntoDiagnostic;
@@ -27,6 +28,7 @@ use stdin::StdinEvent;
 
 mod stdout;
 use stdout::GhciStdout;
+use stdout::StdoutEvent;
 
 mod stderr;
 use stderr::GhciStderr;
@@ -34,6 +36,7 @@ use stderr::GhciStderr;
 mod show_modules;
 use show_modules::ModuleSet;
 
+use crate::aho_corasick::AhoCorasickExt;
 use crate::buffers::LINE_BUFFER_CAPACITY;
 use crate::event_filter::FileEvent;
 use crate::incremental_reader::IncrementalReader;
@@ -131,8 +134,6 @@ impl Ghci {
             error_path: error_path.clone(),
         }));
 
-        let (init_sender, init_receiver) = oneshot::channel::<()>();
-
         // Three tasks for my three beautiful streams.
         let stdout = task::spawn(
             GhciStdout {
@@ -142,8 +143,9 @@ impl Ghci {
                 stderr_sender,
                 receiver: stdout_receiver,
                 buffer: vec![0; LINE_BUFFER_CAPACITY],
+                prompt_patterns: AhoCorasick::from_anchored_patterns([PROMPT]),
             }
-            .run(init_sender),
+            .run(),
         );
         let stderr = task::spawn(
             GhciStderr {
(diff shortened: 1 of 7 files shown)


src/ghci/mod.rs
```diff
@@ -174,20 +176,24 @@ impl Ghci {
         };
 
         // Wait for the stdout job to start up.
-        init_receiver.await.into_diagnostic()?;
```
9999years, line 177: No functional changes here, just some reorganization.

src/ghci/stdout.rs
```diff
@@ -36,12 +39,56 @@ pub struct GhciStdout {
...
+            }
+
+            tracing::debug!("Waiting {duration:?} before retrying");
+            tokio::time::sleep(duration).await;
+        }
+
+        Ok(())
+    }
+
+    async fn run_inner(&mut self) -> miette::Result<()> {
```
9999years, line 69: This isn't new, I just moved a bunch of the startup logic into the `initialize` method and it confused the diff.

src/main.rs
```diff
@@ -19,6 +19,7 @@ use tokio::sync::Mutex;
 
 #[tokio::main]
 async fn main() -> miette::Result<()> {
+    miette::set_panic_hook();
```
9999years, line 22: Print panics nicely with our error library. I might try to replace most uses of `miette` with `color-eyre` at some point because I like its output formatting better.

Gabriella439 (approved):

PR #11, opened 2023-08-09
Run tests on save

9999years:
This adds `--test` and `--setup` arguments which can be used to run a test suite on reload.

### DIFF
--- a/src/ghci/mod.rs
+++ b/src/ghci/mod.rs
@@ -1,10 +1,12 @@
 //! The core [`Ghci`] session struct.
 
+use std::collections::BTreeMap;
 use std::fmt::Debug;
 use std::fmt::Display;
 use std::process::Stdio;
 use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
+use std::time::Instant;
 
 use aho_corasick::AhoCorasick;
 use camino::Utf8PathBuf;
@@ -74,6 +76,10 @@ pub struct Ghci {
     modules: ModuleSet,
     /// Path to write errors to, if any. Like `ghcid.txt`.
     error_path: Option<Utf8PathBuf>,
+    /// `ghci` commands to run on startup.
+    setup_commands: Vec<String>,
+    /// `ghci` command to run tests.
+    test_command: Option<String>,
 }
 
 impl Debug for Ghci {
@@ -91,7 +97,11 @@ impl Ghci {
     pub async fn new(
         command_arc: Arc<Mutex<Command>>,
         error_path: Option<Utf8PathBuf>,
+        setup_commands: Vec<String>,
+        test_command: Option<String>,
     ) -> miette::Result<Arc<Mutex<Self>>> {
+        let start_instant = Instant::now();
+
         let mut child = {
             let mut command = command_arc.lock().await;
 
@@ -132,36 +142,42 @@ impl Ghci {
             sync_count: AtomicUsize::new(0),
             modules: Default::default(),
             error_path: error_path.clone(),
+            setup_commands: setup_commands.clone(),
+            test_command,
         }));
 
         // Three tasks for my three beautiful streams.
         let stdout = task::spawn(
             GhciStdout {
-                ghci: Arc::downgrade(&ret),
                 reader: IncrementalReader::new(stdout).with_writer(tokio::io::stdout()),
-                stdin_sender: stdin_sender.clone(),
-                stderr_sender,
+                stderr_sender: stderr_sender.clone(),
                 receiver: stdout_receiver,
                 buffer: vec![0; LINE_BUFFER_CAPACITY],
                 prompt_patterns: AhoCorasick::from_anchored_patterns([PROMPT]),
+                mode: Mode::Compiling,
             }
             .run(),
         );
         let stderr = task::spawn(
             GhciStderr {
-                ghci: Arc::downgrade(&ret),
                 reader: BufReader::new(stderr).lines(),
                 receiver: stderr_receiver,
-                buffer: String::with_capacity(LINE_BUFFER_CAPACITY),
+                compilation_summary: String::new(),
+                buffers: BTreeMap::from([
+                    (Mode::Compiling, String::with_capacity(LINE_BUFFER_CAPACITY)),
+                    (Mode::Testing, String::with_capacity(LINE_BUFFER_CAPACITY)),
+                ]),
                 error_path,
+                mode: Mode::Compiling,
+                has_unwritten_data: false,
             }
             .run(),
         );
         let stdin = task::spawn(
             GhciStdin {
-                ghci: Arc::downgrade(&ret),
                 stdin,
                 stdout_sender: stdout_sender.clone(),
+                stderr_sender: stderr_sender.clone(),
                 receiver: stdin_receiver,
             }
             .run(),
@@ -177,6 +193,8 @@ impl Ghci {
 
         // Wait for the stdout job to start up.
         {
+            let span = tracing::debug_span!("Stdout startup");
+            let _enter = span.enter();
             let (sender, receiver) = oneshot::channel();
             stdout_sender
                 .send(StdoutEvent::Initialize(sender))
@@ -187,15 +205,22 @@ impl Ghci {
 
         // Perform start-of-session initialization.
         {
+            let span = tracing::debug_span!("Start-of-session initialization");
+            let _enter = span.enter();
             let (sender, receiver) = oneshot::channel();
             stdin_sender
-                .send(StdinEvent::Initialize(sender))
+                .send(StdinEvent::Initialize {
+                    sender,
+                    setup_commands,
+                })
                 .await
                 .into_diagnostic()?;
             receiver.await.into_diagnostic()?;
         }
 
         {
+            let span = tracing::debug_span!("Start-of-session sync");
+            let _enter = span.enter();
             // Sync up for any prompts.
             let mut guard = ret.lock().await;
             guard.sync().await?;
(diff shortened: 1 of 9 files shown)


src/ghci/mod.rs
```diff
@@ -177,6 +193,8 @@ impl Ghci {
 
         // Wait for the stdout job to start up.
         {
+            let span = tracing::debug_span!("Stdout startup");
+            let _enter = span.enter();
```
9999years, line 197: Added some span markers to the startup process to help with debugging.

src/ghci/mod.rs
```diff
@@ -262,9 +291,17 @@ impl Ghci {
...
+            return Self::new(
+                command,
+                guard.error_path.clone(),
+                guard.setup_commands.clone(),
+                guard.test_command.clone(),
+            )
+            .await;
         }
 
+        let needs_add_or_reload = !add.is_empty() || !needs_reload.is_empty();
```
9999years, line 303: This needs to be up here to avoid a borrow error.

src/ghci/mod.rs
```diff
@@ -281,15 +318,31 @@ impl Ghci {
...
                 .stdin_channel
                 .send(StdinEvent::Reload(sender))
                 .await
                 .into_diagnostic()?;
             receiver.await.into_diagnostic()?;
         }
 
+        if needs_add_or_reload {
+            // If we loaded or reloaded any modules, we should run tests.
+            // TODO: Skip this if there were compilation failures?
```
9999years, line 332: I should implement this before merge.

src/ghci/mod.rs
```diff
@@ -356,6 +424,40 @@ impl Ghci {
     }
 }
 
+/// The mode a `ghci` session is in. This is used to track output, particularly for the error log
+/// (`ghcid.txt`).
+///
+/// Note: The [`Ord`] implementation on this type determines the order in which sections will be
+/// written to the error log.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum Mode {
```
9999years, line 433: New concept! The `GhciStdin` task keeps track of the mode (because the input it sends over stdin is what causes `ghci` to do different things...) and sends messages to the other tasks when the mode needs to be updated.

src/ghci/stderr.rs
```diff
@@ -12,24 +12,47 @@ use tokio::io::Lines;
...
+
+    /// Set the writer's mode.
+    Mode {
+        mode: Mode,
+        sender: oneshot::Sender<()>,
+    },
 }
 
 pub struct GhciStderr {
-    pub ghci: Weak<Mutex<Ghci>>,
```
9999years, line 28: These pointers weren't being used for anything and add complexity so I got rid of them.

src/ghci/stdin.rs
```diff
@@ -122,41 +144,78 @@ impl GhciStdin {
     }
 
     #[instrument(skip_all, level = "debug")]
-    async fn initialize(&mut self, sender: oneshot::Sender<()>) -> miette::Result<()> {
-        // For Stack, send a blank line to skip an initial prompt.
-        //
-        // See: https://github.com/ndmitchell/ghcid/issues/57
-        self.stdin.write_all(b"\n").await.into_diagnostic()?;
```
9999years, line 129: This caused a deadlock/race condition... sometimes the stdout task would've caught up with the prompt in time, sometimes it got stuck waiting for a prompt that never came. We can implement this again later (and I (filed a ticket for it)[https://linear.app/mercury/issue/DUX-1292/multi-project-stack-support]).

src/ghci/stdin.rs
```diff
@@ -122,41 +144,78 @@ impl GhciStdin {
...
+        Ok(())
+    }
+
     #[instrument(skip(self), level = "debug")]
     async fn add_module(
         &mut self,
         path: Utf8PathBuf,
         sender: oneshot::Sender<()>,
     ) -> miette::Result<()> {
-        // We use `:load` here instead of `:add` because `:add` forces interpreted mode.
```
9999years, line 152: Not sure where I got this from but it's disastrously untrue. Fixed now.

src/ghci/stdin.rs
```diff
@@ -185,4 +246,34 @@ impl GhciStdin {
             .into_diagnostic()?;
         Ok(())
     }
+
+    #[instrument(skip(self), level = "debug")]
+    async fn set_mode(&self, mode: Mode) -> miette::Result<()> {
```
9999years, line 251: Here's where we inform the other tasks of a new mode.

src/ghci/stdout.rs
```diff
@@ -99,23 +117,16 @@ impl GhciStdout {
...
-            .reader
-            .try_read_until(
-                &init_prompt_patterns,
-                WriteBehavior::NoFinalLine,
-                &mut self.buffer,
-            )
-            .await?
-        {
-            tracing::debug!(?lines, "initial ghci prompt");
-        }
```
9999years, line 118: Turns out `try_read_until` would _sometimes_ do a blocking read call and get stuck forever here.

src/ghci/stdout.rs
```diff
@@ -183,4 +201,33 @@ impl GhciStdout {
...
+    }
+
+    async fn get_status_from_compile_output(&mut self, mut lines: Lines) -> miette::Result<()> {
+        // TODO: This would be pretty clumsy if we wanted to use this regex in multiple places.
+        // Might be worth bringing in `lazy_static`.
+        let compilation_finished_re = COMPILATION_FINISHED_RE
+            .get_or_init(|| Regex::new(r"^(?:Ok|Failed), [0-9]+ modules loaded.$").unwrap());
+
+        if let Some(line) = lines.pop() {
+            if compilation_finished_re.is_match(&line) {
```
9999years, line 218: Here I should inform the `Ghci` session that compilation succeeded or failed so it can use that information to determine if tests should be run.

src/ghci/stdout.rs
```diff
@@ -183,4 +201,33 @@ impl GhciStdout {
...
+    async fn set_mode(&mut self, sender: oneshot::Sender<()>, mode: Mode) {
+        self.mode = mode;
+        let _ = sender.send(());
+    }
+
+    async fn get_status_from_compile_output(&mut self, mut lines: Lines) -> miette::Result<()> {
+        // TODO: This would be pretty clumsy if we wanted to use this regex in multiple places.
+        // Might be worth bringing in `lazy_static`.
+        let compilation_finished_re = COMPILATION_FINISHED_RE
+            .get_or_init(|| Regex::new(r"^(?:Ok|Failed), [0-9]+ modules loaded.$").unwrap());
```
evanrelf, line 215: Or you could use a function with a static.

```rust
fn compilation_finished_re() -> Regex {
    static RE: OnceCell<Regex> = OnceCell::new();
    RE.get_or_init(|| Regex::new(r"^(?:Ok|Failed), [0-9]+ modules loaded.$").unwrap())
}

```

9999years: Good idea, implemented.

9999years:
* main
  * **PR #11** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/11" target="_blank">Graphite</a>  👈
    * **PR #30** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/30" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/11?utm_source=stack-comment).

src/ghci/stdin.rs
```diff
@@ -8,37 +8,50 @@ use tokio::io::AsyncWriteExt;
...
-    Initialize(oneshot::Sender<()>),
+    Initialize {
+        sender: oneshot::Sender<()>,
+        setup_commands: Vec<String>,
+    },
     /// Reload the `ghci` session with `:reload`.
-    Reload(oneshot::Sender<()>),
-    /// Add a module to the `ghci` session by path with `:load`.
-    AddModule(Utf8PathBuf, oneshot::Sender<()>),
+    Reload(oneshot::Sender<Option<Result<(), ()>>>),
```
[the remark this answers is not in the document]
9999years: Fixed, though I'm not sure it makes the code much better.

src/ghci/stdout.rs
```diff
@@ -99,31 +116,24 @@ impl GhciStdout {
...
-                &init_prompt_patterns,
-                WriteBehavior::NoFinalLine,
-                &mut self.buffer,
-            )
-            .await?
-        {
-            tracing::debug!(?lines, "initial ghci prompt");
-        }
+        let init_prompt_patterns = AhoCorasick::from_anchored_patterns(["ghci> "]);
+        let (sender, receiver) = oneshot::channel();
```
Gabriella439, line 123: What is the purpose of this extra channel?  The reason I ask is because, looking at the `prompt` method's definition it seems like the `sender` channel here isn't passed off to any other thread and is only used to receive the result of the `prompt` function, but then I'm wondering if the `prompt` could just return that result instead of passing it back via this channel?

9999years: Usually we're calling the `prompt` function from a different thread; the general flow is the `GhciStdin` job writes a line to stdin and then sends a `Prompt` message to this channel, which waits until it sees a prompt on stdout, and then notifies the sender that it saw a prompt. (This is because sending a message to `GhciStdout` will usually complete immediately, but the prompt will only appear later, so we need a synchronization mechanism.)

For this particular call, we could just call the method, but it needs the `sender` parameter for the majority of use-cases.

This channel also isn't entirely new, it was just moved from the `Ghci::new` method in `src/ghci/mod.rs` to here.

PR #29, opened 2023-08-10
Scripted mode

9999years:
Adds `--script` and `--script-file` to run `ghcid-ng` in scripted mode: a series of `ghci` commands are executed and then it exits.

I intend this to be used first for writing integration tests — right now, `ghcid-ng` is largely untested. In service of that, I'd also like to introduce a "special" command `:ghcid-ng wait [timeout]` that can be used as a command in a `--script` or `--script-file`. Instead of being executed as a `ghci` command like the others, it would signal `ghcid-ng` to wait for a file to change, perform a reload in response, and then continue.


9999years:
-> #30

evanrelf:
For future reference, GitHub lets you change the base branch after creating a PR:

image

I think that's why this got closed.

PR #30, opened 2023-08-10
Add server mode

9999years:
Adds `--listen` and `--notify` to run `ghcid-ng` in server mode.

I intend this to be used first for writing integration tests — right now, `ghcid-ng` is largely untested.

This was originally going to be a "scripted mode" which read commands from a file, but a server architecture seems more broadly useful. @evanrelf also pointed out that this could be used to implement test-on-demand or reload-on-demand features; you could imagine whacking a keystroke in your editor to fire a test run, and getting notified of the results inline.

This interlocks with a more long-term desire to expose the `ghci` prompt to users while reloads aren't running. It would be super useful to be able to run arbitrary commands, set breakpoints, etc. (Perhaps `ghcid-ng` could even expose a [debug adapter protocol](https://microsoft.github.io/debug-adapter-protocol/) at some point!)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,7 +183,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6798148dccfbff0fae41c7574d2fa8f1ef3492fba0face179de5d8d447d67b05"
 dependencies = [
  "memchr",
- "regex-automata 0.3.4",
+ "regex-automata 0.3.6",
  "serde",
 ]
 
@@ -558,6 +558,7 @@ dependencies = [
  "once_cell",
  "owo-colors",
  "pretty_assertions",
+ "regex",
  "shell-words",
  "strip-ansi-escapes",
  "tap",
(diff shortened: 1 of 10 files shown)


9999years:
* main
  * **PR #11** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/11" target="_blank">Graphite</a> 
    * **PR #30** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/30" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/30?utm_source=stack-comment).

evanrelf:
I wonder if you could use [vhs](https://github.com/charmbracelet/vhs) instead

9999years:
> I wonder if you could use [vhs](https://github.com/charmbracelet/vhs) instead. Edit: hmm not sure it allows arbitrary file output to be checked by a test harness, it might just output images.

I want to load `ghcid-ng`, edit a file (I can't find a way to do this with vhs), and confirm that reloading works. (I want to do a number of variations on this, so I can check things like running tests, `ghcid.txt` output, etc.) Ideally I'd like to do this without specifying a bunch of timeouts and praying.

evanrelf:
[expect](https://core.tcl-lang.org/expect/index) is what I was trying to remember for this kinda thing!

Also kinda unrelated, the `expect` package provides an `unbuffer` command that can help with preserving colored output. I think it makes the command think its outputting to a terminal. For example:

```ShellSession
$ unbuffer cabal build |& tee cabal-build.log # or whatever you pipe to, I dunno
```

`unbuffer` might be useful if you don't wanna use `-fdiagnostics-color=always`, or there's some other program that doesn't have an equivalent `--color=force` option.

PR #32, opened 2023-08-17
Don't run tests when compilation fails

9999years:
This was the intended behavior in #11, but I got the booleans backwards. Don't worry, tests are next on the roadmap...

### DIFF
--- a/src/ghci/mod.rs
+++ b/src/ghci/mod.rs
@@ -303,7 +303,7 @@ impl Ghci {
         }
 
         let needs_add_or_reload = !add.is_empty() || !needs_reload.is_empty();
-        let mut compilation_failed = needs_add_or_reload;
+        let mut compilation_failed = false;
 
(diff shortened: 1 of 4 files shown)


src/ghci/mod.rs
```diff
@@ -313,7 +313,7 @@ impl Ghci {
             for path in add {
                 let add_result = this.lock().await.add_module(path).await?;
                 if let Some(CompilationResult::Err) = add_result {
-                    compilation_failed = false;
+                    compilation_failed = true;
```
9999years, line 316: Oops

Gabriella439: oh yeah, i forgot about this, too.  in fact, this was the line of code that made me ask about the result type refactor because i wasn't sure if i understood the type correctly

Gabriella439 (approved):

PR #34, opened 2023-08-17
Add server mode

9999years:
Adds the ability run `ghcid-ng` in server mode.

I intend this to be used first for writing integration tests — right now, `ghcid-ng` is largely untested.

This was originally going to be a "scripted mode" which read commands from a file, but a server architecture seems more broadly useful. @evanrelf also pointed out that this could be used to implement test-on-demand or reload-on-demand features; you could imagine whacking a keystroke in your editor to fire a test run, and getting notified of the results inline.

This interlocks with a more long-term desire to expose the `ghci` prompt to users while reloads aren't running. It would be super useful to be able to run arbitrary commands, set breakpoints, etc. (Perhaps `ghcid-ng` could even expose a [debug adapter protocol](https://microsoft.github.io/debug-adapter-protocol/) at some point!)


9999years:
* main
  * **PR #34** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/34" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/34?utm_source=stack-comment).

PR #39, opened 2023-08-22
Move `Ghci` out of the `Arc<Mutex<_>>`

9999years:
In #11 I removed the `Weak<Ghci>` pointers from the ghci task structs, favoring the `mpsc` channels for communication over locking a mutex. This means I can make several simplifications, in particular turning `Ghci::reload` into a real method that uses `std::mem::replace`.

PR #40, opened 2023-08-22
Allow customizing logged span events

9999years:
Currently, new/close span events are logged (that is, when a span is created for the first time, usually when an `#[instrument]`ed function is called, and when spans are closed, usually when those functions return). It can be useful to disable those for terser logs, or to enable more events for greater detail.

This change adds a `--trace-spans` argument to customize which span events are logged.

Before shipping 1.0, we should clean up the option names. Maybe `--tracing-filter` could be `--log` or something.

### DIFF
--- a/src/clap/fmt_span.rs
+++ b/src/clap/fmt_span.rs
@@ -1,0 +1,62 @@
+//! Adapter for pasing the [`FmtSpan`] type.
+
+use clap::builder::EnumValueParser;
+use clap::builder::PossibleValue;
+use clap::builder::TypedValueParser;
+use clap::builder::ValueParserFactory;
+use tracing_subscriber::fmt::format::FmtSpan;
+
+/// Wrapper around [`FmtSpan`].
+#[derive(Clone)]
+pub struct FmtSpanWrapper(FmtSpan);
+
+impl From<FmtSpanWrapper> for FmtSpan {
+    fn from(value: FmtSpanWrapper) -> Self {
+        value.0
+    }
+}
+
+impl clap::ValueEnum for FmtSpanWrapper {
+    fn value_variants<'a>() -> &'a [Self] {
+        &[
+            Self(FmtSpan::NEW),
+            Self(FmtSpan::ENTER),
+            Self(FmtSpan::EXIT),
+            Self(FmtSpan::CLOSE),
+            Self(FmtSpan::NONE),
+            Self(FmtSpan::ACTIVE),
+            Self(FmtSpan::FULL),
+        ]
+    }
+
+    fn to_possible_value(&self) -> Option<PossibleValue> {
+        Some(match self.0 {
+            FmtSpan::NEW => PossibleValue::new("new").help("Log when spans are created"),
+            FmtSpan::ENTER => PossibleValue::new("enter").help("Log when spans are entered"),
+            FmtSpan::EXIT => PossibleValue::new("exit").help("Log when spans are exited"),
+            FmtSpan::CLOSE => PossibleValue::new("close").help("Log when spans are dropped"),
+            FmtSpan::NONE => PossibleValue::new("none").help("Do not log span events"),
+            FmtSpan::ACTIVE => {
+                PossibleValue::new("active").help("Log when spans are entered/exited")
+            }
+            FmtSpan::FULL => PossibleValue::new("full").help("Log all span events"),
+            _ => {
+                return None;
+            }
+        })
+    }
+}
+
+/// [`clap`] parser factory for [`FmtSpan`] values.
+pub struct FmtSpanParserFactory;
+
+impl ValueParserFactory for FmtSpanParserFactory {
+    type Parser = clap::builder::MapValueParser<
+        EnumValueParser<FmtSpanWrapper>,
+        fn(FmtSpanWrapper) -> FmtSpan,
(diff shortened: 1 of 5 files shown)


9999years:
* main
  * **PR #39** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/39" target="_blank">Graphite</a> 
    * **PR #40** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/40" target="_blank">Graphite</a>  👈
      * **PR #34** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/34" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/40?utm_source=stack-comment).

src/cli.rs
```diff
@@ -91,6 +94,29 @@ pub struct LoggingOpts {
...
+#[derive(Debug, Clone, clap::Args)]
+#[clap(next_help_heading = "Server options")]
+pub struct ServerOpts {
+    /// Start in server mode, binding to the given socket path.
+    ///
+    /// The socket can be used to send commands to `ghcid-ng` and to receive event notifications
+    /// back in turn. The communication protocol is unstable for now.
+    #[arg(long, value_name = "PATH")]
+    pub socket: Option<Utf8PathBuf>,
 }
```
evanrelf, line 120: I don't think `ServerOpts` belongs in this PR.

src/clap/fmt_span.rs
```diff
@@ -0,0 +1,62 @@
...
+        &[
+            Self(FmtSpan::NEW),
+            Self(FmtSpan::ENTER),
+            Self(FmtSpan::EXIT),
+            Self(FmtSpan::CLOSE),
+            Self(FmtSpan::NONE),
+            Self(FmtSpan::ACTIVE),
+            Self(FmtSpan::FULL),
+        ]
+    }
```
evanrelf, line 30: 1. I'm surprised this lifetime can't be elided?

2. Could this list be generated with the Rust equivalent of `[minBound .. maxBound]` in Haskell? The compiler won't tell you if a new variant is added, making this list incomplete.

src/clap/fmt_span.rs
```diff
@@ -0,0 +1,62 @@
...
+            FmtSpan::CLOSE => PossibleValue::new("close").help("Log when spans are dropped"),
+            FmtSpan::NONE => PossibleValue::new("none").help("Do not log span events"),
+            FmtSpan::ACTIVE => {
+                PossibleValue::new("active").help("Log when spans are entered/exited")
+            }
+            FmtSpan::FULL => PossibleValue::new("full").help("Log all span events"),
+            _ => {
+                return None;
+            }
+        })
```
evanrelf, line 46: Is this one of those permanently non-exhaustive enum things, or are you ignoring some of the options? If the latter, why?

evanrelf (approved):

src/cli.rs
9999years: oops

evanrelf (approved):

src/clap/fmt_span.rs
9999years: 1. Me too! I guess the trait definition has an explicit lifetime so you're not allowed to elide it here?

<details><summary>Compilation error when I elide the lifetime</summary>


```
   error[E0106]: missing lifetime specifier
  --> src/clap/fmt_span.rs:20:28
   |
20 |     fn value_variants() -> &[Self] {
   |                            ^ expected named lifetime parameter
   |
   = help: this function's return type contains a borrowed value, but there is no value for it to be borrowed from
help: consider using the `'static` lifetime
   |
20 |     fn value_variants() -> &'static [Self] {
   |                             +++++++

error[E0195]: lifetime parameters or bounds on method `value_variants` do not match the trait declaration
  --> src/clap/fmt_span.rs:20:22
   |
20 |     fn value_variants() -> &[Self] {
   |                      ^ lifetimes do not match method in trait

Some errors have detailed explanations: E0106, E0195.
For more information about an error, try `rustc --explain E0106`.
```

</details>

2. I *wish!* This has been [an open issue since before Rust 1.0](https://github.com/rust-lang/rfcs/issues/284#issuecomment-1592343574) (peep the Ruby-like syntax in the [original issue here](https://github.com/rust-lang/rust/issues/5417)). There's a few crates that implement this functionality for `enum` types with `#[derive]` macros, but they're pretty heavy so we don't use them here.

src/clap/fmt_span.rs
9999years: The former. The `FmtSpan` type is actually defined as a wrapper around a `u8`, not an `enum`:

```
pub struct FmtSpan(u8);

impl FmtSpan {
    pub const NEW: FmtSpan = FmtSpan(1 << 0);
    // ...
}
```

https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.FmtSpan.html

So these _look_ like `enum` variants, but they're actually constants in the type's namespace. (You can tell something funky is up because they're written in `ALL_CAPS` rather than `TitleCase`.)

In short: this is necessary to make the compiler not complain.

PR #45, opened 2023-08-24
Remove `GhciStdin` process

Gabriella439:
This removes the process that manages `GhciStdin` and instead replaces the handle to that process with the `GhciStdin` struct.

The motivation behind this change is to remove the need for a `StdinEvent` type and avoid the need to spawn and monitor/restart a process.  This also simplifies method invocations on that process, too.

### DIFF
--- a/src/ghci/mod.rs
+++ b/src/ghci/mod.rs
@@ -64,8 +64,6 @@ pub struct Ghci {
     stdout: JoinHandle<miette::Result<()>>,
     /// The handle for the stderr reader task.
     stderr: JoinHandle<miette::Result<()>>,
-    /// The handle for the stdin interaction task.
-    stdin: JoinHandle<miette::Result<()>>,
     ghci_stdin: GhciStdin,
     /// Count of 'sync' events sent. This lets us sync stdin/stdout -- we write a message to stdin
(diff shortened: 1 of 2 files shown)


src/ghci/mod.rs
```diff
@@ -132,21 +129,28 @@ impl Ghci {
...
+                stdout_sender: stdout_sender.clone(),
+                stderr_sender: stderr_sender.clone(),
+              };
+
         let mut ret = Ghci {
             command: command_arc,
             process: child,
             stdout: stdout_handle,
             stderr: stderr_handle,
             stdin: stdin_handle,
```
9999years, line 144: This handle is no longer associated with a task, so the field should be removed.

9999years: `ghci_stdin` is also a little bit of a clumsy name (this is the `Ghci` struct, what else would it be `stdin` for?), so maybe we should rename `stdout` and `stderr` to `stdout_handle` and `stderr_handle` (to reflect the fact that they're `JoinHandle`s for tasks) and then `ghci_stdin` can just be `stdin`.

9999years (changes requested):
Looks good! This is a great simplification. Let's fix up the `Ghci` fields and then this looks good to merge.

9999years (approved):

PR #49, opened 2023-08-25
Fewer processes and (almost) no `Sender`s

Gabriella439:
The main motivation for this change was to remove `Sender`-based result-passing in favor of passing values directly through return values.  However, in order to do this I had to remove the subprocesses that were managing `GhciStdout` and `GhciStderr`.

The basic approach I used to make this work and appease the borrow-checker was:

- Make the `Ghci` struct own all three of `stdin`/`stdout`/`stderr`
- Any method that required more than one of them would borrow the other ones it was missing

I think there's actually a better way to do this that doesn't require as much parameter passing, which is to move the methods on `GhciStdin`/`GhciStdout`/`GhciStderr` to become methods on `Ghci`, but I'll save that for a separate refactor since this change was already starting to get a big large.  I picked this approach for my first pass since it led to the smallest diff for ease of review.

Then once I finished that refactor I was able get rid of all the `Sender`-related logic and instead pass the results that would have normally been passed via `Sender`s via the function's return value instead.  If you squint you can sort of think of this as analogous to converting code from "continuation-passing style" to "direct style".


src/ghci/stderr.rs
```diff
@@ -56,63 +32,8 @@ pub struct GhciStderr {
...
-            // processed after we write the error log?
-            tokio::select! {
-                Ok(Some(line)) = self.reader.next_line() => {
-                    self.ingest_line(line).await;
-                }
-                Some(event) = self.receiver.recv() => {
-                    self.dispatch(event).await?;
-                }
-            }
-        }
```
9999years, line 93: This bit of code waits for lines to be printed on `ghci`'s stderr and ingests them (puts them in a buffer keyed by the current `Mode`). It also waits for events to be received on the input channel. After this refactor, there is no longer any code that calls `ingest_line`, so stderr output is not printed on stderr (similarly, `ghcid.txt` includes no stderr output, just a `Failed, 2521 modules loaded.` line).

Gabriella439: Alright, I ended up fixing this by keeping the dispatch loop and use of `Sender`s for `stderr.rs`, but left the other stuff mostly the same.

src/ghci/stdout.rs
```diff
@@ -203,20 +128,17 @@ impl GhciStdout {
...
-        let map = ModuleSet::from_lines(&lines)?;
-        let _ = sender.send(map);
-        Ok(())
+        ModuleSet::from_lines(&lines)
     }
 
-    #[instrument(skip(self, sender), level = "debug")]
-    async fn set_mode(&mut self, sender: oneshot::Sender<()>, mode: Mode) {
+    #[instrument(skip(self), level = "debug")]
+    pub async fn set_mode(&mut self, mode: Mode) {
```
9999years, line 140:
```suggestion
    pub fn set_mode(&mut self, mode: Mode) {
```

Can be synchronous.

src/ghci/stdout.rs
```diff
@@ -203,20 +128,17 @@ impl GhciStdout {
     }
 
     #[instrument(skip_all, level = "debug")]
-    async fn show_modules(&mut self, sender: oneshot::Sender<ModuleSet>) -> miette::Result<()> {
+    pub async fn show_modules(&mut self) -> miette::Result<ModuleSet> {
```
9999years, line 131:
```suggestion
    pub fn show_modules(&mut self) -> miette::Result<ModuleSet> {
```

Can be synchronous.

9999years (changes requested):
Looks good. I spotted a couple methods that can be made non-`async`, but after that we can merge.

src/ghci/stdout.rs
Gabriella439: I think this `async` is still necessary.  There's a `.await` in this function

PR #55, opened 2023-08-28
Move to multi-crate workspace

9999years:
Moving to a multi-crate workspace lays the foundation for a `#[test]`-attribute-based test harness. We could also potentially separate the CLI and "library" portions of `ghcid-ng` into separate crates.


flake.nix
```diff
@@ -47,6 +47,10 @@
               pkgs.libiconv
               pkgs.darwin.apple_sdk.frameworks.CoreServices
             ];
+
+            cargoBuildCommand = "cargoWithProfile build --all";
+            cargoCheckExtraArgs = "--all";
+            cargoTestExtraArgs = "--all";
```
9999years, line 53: We need to add `--all` in a few places to get all the packages building/tested.

ghcid-ng/Cargo.toml
```diff
@@ -0,0 +1,56 @@
+[package]
```
9999years, line 1: Unchanged, rename-only.

Gabriella439 (approved):

PR #74, opened 2023-09-11
Rewrite `:show modules` parser with `winnow`

9999years:
This is a small change on the way to the general GHC diagnostics parser.


ghcid-ng/src/ghci/parse/haskell_grammar.rs
```diff
@@ -0,0 +1,74 @@
...
+    separated1::<_, _, (), _, _, _, _>(constructor_name, ".")
+        .recognize()
+        .parse_next(input)
+}
+
+/// A Haskell constructor name.
+///
+/// See: `conid` in <https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-180002.4>
+fn constructor_name<'i>(input: &mut &'i str) -> PResult<&'i str> {
+    // TODO: Support Unicode letters.
```
9999years, line 27: We can do this with the nice `unicode-properties` crate, but I'm not sure we'll need it?

https://docs.rs/unicode-properties/latest/unicode_properties/

ghcid-ng/src/ghci/parse/module_set.rs
```diff
@@ -0,0 +1,81 @@
...
+use miette::Context;
+use miette::IntoDiagnostic;
+
+use super::Module;
+
+/// A collection of source paths, retaining information about loaded modules in a `ghci`
+/// session.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ModuleSet {
+    map: HashSet<Utf8PathBuf>,
```
Gabriella439, line 14: maybe call this field `set`?

ghcid-ng/src/ghci/parse/module_set.rs
```diff
@@ -0,0 +1,81 @@
...
+    #[allow(dead_code)]
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Remove all entries from this set, leaving it empty.
+    #[allow(dead_code)]
+    pub fn clear(&mut self) {
+        self.map.clear();
+    }
```
Gabriella439, line 47: i feel like it would be easier to make the `map` field public. i don't think much is gained from information hiding here and then you wouldn't need these methods

Gabriella439 (approved):

ghcid-ng/src/ghci/parse/module_set.rs
9999years: Oh, good idea (a long time ago, this value was a map).

ghcid-ng/src/ghci/parse/module_set.rs
9999years: Not just information hiding, we don't want users to insert things into the set without knowing about it -- in particular, we want all of our paths to be nicely canonicalized.

Gabriella439: ah okay, never mind then

9999years: I should eventually write a `CanonicalizedUtf8PathBuf` type, though.

Gabriella439 (approved):

PR #75, opened 2023-09-11
Make `IncrementalReader` return strings, not lists of lines

9999years:
This is conceptually somewhat simpler and also gives us output that's easier to feed to parsers.

### DIFF
--- a/ghcid-ng/src/ghci/stdout.rs
+++ b/ghcid-ng/src/ghci/stdout.rs
@@ -13,7 +13,6 @@ use tracing::instrument;
 use crate::aho_corasick::AhoCorasickExt;
 use crate::incremental_reader::IncrementalReader;
 use crate::incremental_reader::WriteBehavior;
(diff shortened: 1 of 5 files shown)


ghcid-ng/src/ghci/stdout.rs
```diff
@@ -69,20 +68,20 @@ impl GhciStdout {
...
             )
             .await?;
-        tracing::debug!(lines = lines.len(), "Got data from ghci");
+        tracing::debug!(bytes = data.len(), "Got data from ghci");
 
         let mut result = None;
         if self.mode == Mode::Compiling {
             result = self
-                .get_status_from_compile_output(lines)
+                .get_status_from_compile_output(data.lines().last())
```
9999years, line 84: In the next PR, I'll replace this with the GHC output parser to avoid this inefficient traversal, but this makes tests pass.

Gabriella439 (approved):

PR #83, opened 2023-09-13
Fix release process

9999years:
The release process was broken when we moved to a Cargo workspace. This fixes it.

- I found out that Cargo workspaces can have a "root crate", which here is `ghcid-ng`. This lets us move all the `ghcid-ng` sources out of the `ghcid-ng` directory and into a top-level `src` directory, which is very nice.

- The inevitable `nix/` directory has been created.

    - Moved a few helper scripts to `nix/` and packaged them.

- The versioning and release process has been moved into Nix as well. This makes everything much slower but at least we know what versions of the tools we're using?
    - At least on Linux. Cross-compilation on macOS with Nix doesn't work: https://github.com/NixOS/nixpkgs/issues/180771


9999years:
* main
  * **PR #83** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/83" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/83?utm_source=stack-comment).

Gabriella439 (dismissed):

Gabriella439 (approved):

PR #77, opened 2023-09-11
Fix "module defined in multiple files" by parsing GHC diagnostics

9999years:
This parses GHC diagnostics and uses the output to avoid `:add`ing failed modules. See the bug upstream here: https://gitlab.haskell.org/ghc/ghc/-/issues/13254

PR #76, opened 2023-09-11
Split `Ghci::reload` up

9999years:
This moves a bunch of logic into `Ghci::get_reload_actions`.

PR #73, opened 2023-09-11
Add before-start hook to the test harness

9999years:
This lets tests run code to modify the project before starting `ghcid-ng`.

The before-start hook is implemented in terms of a new `GhcidNgBuilder` interface.

PR #86, opened 2023-09-19
Add eval support

9999years:
Adds support for evaluating commands in source files on reload.

This looks for line commands starting with `-- $>` or multiline commands delimited by `{- $>` and `<$ -}`.

This also removes the `failed_modules` mechanism from `Ghci`, which was populated by extracting filenames from error messages, and replaces it with a `targets` `ModuleSet` populated by `:show targets`. This should be more consistent.

Finally, paths will be printed as relative paths when possible.


9999years:
* main
  * **PR #73** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/73" target="_blank">Graphite</a> 
    * **PR #74** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/74" target="_blank">Graphite</a> 
      * **PR #75** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/75" target="_blank">Graphite</a> 
        * **PR #76** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/76" target="_blank">Graphite</a> 
          * **PR #77** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/77" target="_blank">Graphite</a> 
            * **PR #78** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/78" target="_blank">Graphite</a> 
              * **PR #86** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/86" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/86?utm_source=stack-comment).

src/ghci/stdin.rs
```diff
@@ -135,15 +137,64 @@ impl GhciStdin {
...
+        stdout.show_paths().await
+    }
+
+    #[instrument(skip(self, stdout), level = "debug")]
+    pub async fn show_targets(
+        &mut self,
+        stdout: &mut GhciStdout,
+        show_paths: &ShowPaths,
+    ) -> miette::Result<ModuleSet> {
+        self.set_mode(stdout, Mode::Internal).await?;
```
Gabriella439, line 157: as a side note, i think at some point (doesn't have to be in this PR) the mode system should be changed so that the mode is provided as an argument to write commands instead of being set out-of-band and tracked as a mutable variable.   That would prevent quite a few potential bugs due to the wrong mode being set for a given write

src/ghci/stdin.rs
```diff
@@ -135,15 +137,64 @@ impl GhciStdin {
...
+            .await
+            .into_diagnostic()?;
+        stdout.prompt(None).await?;
+
+        for line in command.as_ref().lines() {
+            self.stdin
+                .write_all(line.as_bytes())
+                .await
+                .into_diagnostic()?;
+            self.stdin.write_all(b"\n").await.into_diagnostic()?;
```
Gabriella439, line 187: minor suggestion:

```suggestion
            self.stdin
                .write_all(format!("{line}\n").as_bytes())
                .await
                .into_diagnostic()?;
```

src/ghci/parse/show_paths.rs
```diff
@@ -0,0 +1,240 @@
+use std::collections::HashSet;
+
+use camino::Utf8Path;
+use camino::Utf8PathBuf;
+use itertools::Itertools;
+use miette::miette;
+use winnow::ascii::newline;
+use winnow::ascii::space0;
+use winnow::ascii::space1;
```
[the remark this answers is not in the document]
9999years: Tends to make diffs worse in my experience. I really like keeping one item to one line.

src/haskell_source_file.rs
```diff
@@ -4,6 +4,7 @@ use camino::Utf8Path;
 
 /// File extensions for Haskell source code.
 pub const HASKELL_SOURCE_EXTENSIONS: [&str; 9] = [
+    // NOTE: This should start with `hs` so that iterators try the most common extension first.
```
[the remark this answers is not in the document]
9999years: I actually noticed the ghci manual says it only looks at `.hs` and `.lhs`, so we might want to remove all the other extensions...
https://downloads.haskell.org/ghc/latest/docs/users_guide/ghci.html#modules-vs-filenames

src/ghci/stdin.rs
9999years: I was trying to avoid allocating another string by formatting it... it probably doesn't matter too much. Is doing multiple writes worse than allocating?

src/ghci/stdin.rs
9999years: I agree with setting the mode as part of the write commands, but I think it'll still have to be tracked as a mutable variable because the mode influences the way the stdout and stderr readers process their output. (They need to be able to know what the last mode was, not just when a write happens.)

src/haskell_source_file.rs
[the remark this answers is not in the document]
9999years: I think you're right that `hsboot` isn't valid though, removed that.

Gabriella439 (approved):

Gabriella439 (approved):

PR #97, opened 2023-09-22
Graceful shutdown

9999years:
This PR switches `ghcid-ng` from the `watchexec` crate to using the `notify` crate directly. (This is what `watchexec` uses under the hood.) `watchexec` is very nice, but it's largely designed around the needs of the `watchexec(1)` binary (namely running commands when files change). Using `notify` makes the code much nicer.

One of the things that `watchexec` provides that `notify` doesn't is shutting down when Ctrl-C is pressed. Most of the rest of the changes are about implementing this graceful shutdown procedure.

See [the “Graceful Shutdown” section in the Tokio book](https://tokio.rs/tokio/topics/shutdown) for more details.

Originally I implemented this with [the `tokio-graceful-shutdown` crate](https://docs.rs/tokio-graceful-shutdown/latest/tokio_graceful_shutdown/), but it turned out to be missing some features I wanted. (In particular, performing a graceful shutdown when Ctrl-C is pressed and then performing a hard shutdown if Ctrl-C is pressed again.) Fortunately, implementing the subset of features we need wasn't too hard.

PR #118, opened 2023-10-02
Add matcher combinators

9999years:
These matcher combinators will let us express more complex logging concepts, like:
- Match these two events in any order.
- Match either of these two events.
- Match either of these two events, as long as this other event doesn't match first.

There's also a system of checkpoints where log events are read into the default checkpoint. At any time, you can create another checkpoint with `GhciWatch::checkpoint`. Then, log events will be read into the new checkpoint. Later, you can assert that messages were logged in a particular checkpoint or range of checkpoints.

### DIFF
--- a/test-harness/src/checkpoint.rs
+++ b/test-harness/src/checkpoint.rs
@@ -1,0 +1,104 @@
+use std::fmt::Debug;
+use std::fmt::Display;
+use std::ops::Range;
+use std::ops::RangeFrom;
+use std::ops::RangeFull;
+use std::ops::RangeInclusive;
+use std::ops::RangeTo;
+use std::ops::RangeToInclusive;
+use std::slice::SliceIndex;
+
+use crate::Event;
+
+/// A checkpoint in a `ghciwatch` run.
+///
+/// [`crate::GhciWatch`] provides methods for asserting that events are logged, or waiting for
+/// events to be logged in the future.
+///
+/// To avoid searching thousands of log events for each assertion, and to provide greater
+/// granularity for assertions, you can additionally assert that events are logged between
+/// particular checkpoints.
+///
+/// Checkpoints can be constructed with [`crate::GhciWatch::first_checkpoint`],
+/// [`crate::GhciWatch::current_checkpoint`], and [`crate::GhciWatch::checkpoint`].
+#[derive(Debug, Clone, Copy)]
+pub struct Checkpoint(pub(crate) usize);
+
+impl Checkpoint {
+    /// Convert this checkpoint into an index.
+    pub fn into_index(self) -> usize {
+        self.0
+    }
+}
+
+impl Display for Checkpoint {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// A type that can be used to index a set of checkpoints.
+pub trait CheckpointIndex: Clone {
+    /// The resulting index type.
+    type Index: SliceIndex<[Vec<Event>], Output = [Vec<Event>]> + Debug;
+
+    /// Convert the value into an index.
+    fn into_index(self) -> Self::Index;
+}
+
+impl CheckpointIndex for Checkpoint {
+    type Index = RangeInclusive<usize>;
+
+    fn into_index(self) -> Self::Index {
+        let index = self.into_index();
+        index..=index
+    }
+}
+
+impl CheckpointIndex for Range<Checkpoint> {
+    type Index = Range<usize>;
+
+    fn into_index(self) -> Self::Index {
+        self.start.into_index()..self.end.into_index()
+    }
+}
+
+impl CheckpointIndex for RangeFrom<Checkpoint> {
+    type Index = RangeFrom<usize>;
+
+    fn into_index(self) -> Self::Index {
+        self.start.into_index()..
+    }
+}
+
+impl CheckpointIndex for RangeFull {
+    type Index = RangeFull;
+
+    fn into_index(self) -> Self::Index {
+        self
+    }
+}
+
+impl CheckpointIndex for RangeInclusive<Checkpoint> {
+    type Index = RangeInclusive<usize>;
+
+    fn into_index(self) -> Self::Index {
+        self.start().into_index()..=self.end().into_index()
+    }
+}
+
+impl CheckpointIndex for RangeTo<Checkpoint> {
+    type Index = RangeTo<usize>;
+
+    fn into_index(self) -> Self::Index {
+        ..self.end.into_index()
+    }
+}
+
+impl CheckpointIndex for RangeToInclusive<Checkpoint> {
+    type Index = RangeToInclusive<usize>;
+
+    fn into_index(self) -> Self::Index {
+        ..=self.end.into_index()
+    }
+}
--- a/test-harness/src/checkpoint.rs
+++ b/test-harness/src/checkpoint.rs
@@ -25,8 +25,8 @@ use crate::Event;
 pub struct Checkpoint(pub(crate) usize);
 
 impl Checkpoint {
-    /// Convert this checkpoint into an index.
-    pub fn into_index(self) -> usize {
+    /// Get the underlying `usize` from this checkpoint.
+    pub fn into_inner(self) -> usize {
         self.0
     }
 }
@@ -50,7 +50,7 @@ impl CheckpointIndex for Checkpoint {
     type Index = RangeInclusive<usize>;
 
     fn into_index(self) -> Self::Index {
-        let index = self.into_index();
+        let index = self.into_inner();
         index..=index
     }
 }
@@ -59,7 +59,7 @@ impl CheckpointIndex for Range<Checkpoint> {
     type Index = Range<usize>;
 
     fn into_index(self) -> Self::Index {
-        self.start.into_index()..self.end.into_index()
+        self.start.into_inner()..self.end.into_inner()
     }
 }
 
@@ -67,7 +67,7 @@ impl CheckpointIndex for RangeFrom<Checkpoint> {
     type Index = RangeFrom<usize>;
 
     fn into_index(self) -> Self::Index {
-        self.start.into_index()..
+        self.start.into_inner()..
     }
 }
 
@@ -83,7 +83,7 @@ impl CheckpointIndex for RangeInclusive<Checkpoint> {
     type Index = RangeInclusive<usize>;
 
     fn into_index(self) -> Self::Index {
-        self.start().into_index()..=self.end().into_index()
+        self.start().into_inner()..=self.end().into_inner()
     }
 }
 
@@ -91,7 +91,7 @@ impl CheckpointIndex for RangeTo<Checkpoint> {
     type Index = RangeTo<usize>;
 
     fn into_index(self) -> Self::Index {
-        ..self.end.into_index()
+        ..self.end.into_inner()
     }
 }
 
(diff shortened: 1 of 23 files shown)


9999years:
* main
  * **PR #118** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/118" target="_blank">Graphite</a>  👈
    * **PR #97** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/97" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/MercuryTechnologies/ghcid-ng/118?utm_source=stack-comment).

test-harness/src/checkpoint.rs
```diff
@@ -0,0 +1,104 @@
...
+/// particular checkpoints.
+///
+/// Checkpoints can be constructed with [`crate::GhcidNg::first_checkpoint`],
+/// [`crate::GhcidNg::current_checkpoint`], and [`crate::GhcidNg::checkpoint`].
+#[derive(Debug, Clone, Copy)]
+pub struct Checkpoint(pub(crate) usize);
+
+impl Checkpoint {
+    /// Convert this checkpoint into an index.
+    pub fn into_index(self) -> usize {
```
Gabriella439, line 29: It confused me quite a bit that this `into_index` method on `Checkpoint` had the same name as the `into_index` method on `CheckpointIndex`.  Maybe give this method a different name or just access the `.usize` field directly instead of using this method since that field is already crate-visible

test-harness/src/checkpoint.rs
```diff
@@ -0,0 +1,104 @@
...
+/// events to be logged in the future.
+///
+/// To avoid searching thousands of log events for each assertion, and to provide greater
+/// granularity for assertions, you can additionally assert that events are logged between
+/// particular checkpoints.
+///
+/// Checkpoints can be constructed with [`crate::GhcidNg::first_checkpoint`],
+/// [`crate::GhcidNg::current_checkpoint`], and [`crate::GhcidNg::checkpoint`].
+#[derive(Debug, Clone, Copy)]
+pub struct Checkpoint(pub(crate) usize);
```
Gabriella439, line 25: I personally feel like this `Checkpoint` "newtype" doesn't buy you much over just using `usize` directly and the `Checkpoint` type adds a lot of complexity here.  Most of the `CheckpointIndex` `impl`s don't really do anything other than `coerce` `Checkpoint` to `usize`

9999years: It's a bounds-checked index, so it's guaranteed to be safe to use. If we used a `usize` the user could pass in garbage and cause the test to panic. Probably not that bad, but we _do_ get a safety guarantee out of this.

9999years: Oh, actually, the thing it does give us is a `SliceIndex` implementation with slice output for single checkpoints. This lets us treat single checkpoints like `1` and ranges like `1..3` or `2..` the same. The standard `SliceIndex` implementations instead return a `T` for a `usize` and a `[T]` for ranges of `usize`.

So we'd need to keep the `CheckpointIndex` trait around, even if it was just implemented for `usize` and ranges of `usize`.

test-harness/src/ghciwatch.rs
```diff
@@ -190,35 +232,109 @@ impl GhciWatch {
...
-                            return Ok(event);
-                        } else if let Some(negative_matcher) = &negative_matcher {
-                            if negative_matcher.matches(&event) {
-                                return Err(miette!("Found a log event matching {negative_matcher}"));
-                            }
-                        }
-                    }
+                let event = self.read_event().await?;
+                if matcher.matches(event)? {
+                    return Ok(event.clone());
```
evanrelf, line 337: (struggled to select the exact region I want in GitHub, sorry)

This block reads really strangely to me. I don't think the `Ok(Ok(event))` or `Ok(Err(err))` match arms are reachable. You'll either `return` early from the whole function, or you'll get `Err(_)` from the timeout. Otherwise you're just looping infinitely.

I would either:

1. Change the current `match the_block` to `if the_block.is_err()`
2. `break` from the loop with the matching event value instead of `return`ing from the function

I don't think your current code is incorrect, it's just weird lol.

test-harness/src/ghciwatch.rs
```diff
@@ -227,97 +343,108 @@ impl GhciWatch {
...
 
     /// Wait until `ghciwatch` adds new modules to the `ghci` session.
     pub async fn wait_until_add(&mut self) -> miette::Result<()> {
         // TODO: It would be nice to verify which modules are being added.
-        self.assert_logged("^Adding modules to ghci:\n")
-            .await
-            .map(|_| ())
+        self.wait_for_log(BaseMatcher::ghci_add()).await?;
+        Ok(())
     }
```
evanrelf, line 441: Looks like you have two styles here, and you're almost consistent, but not 100%

1. `foo.await.map(|_| ())`
2. `foo.await?; Ok(())`

test-harness/src/matcher/and_matcher.rs
```diff
@@ -0,0 +1,72 @@
...
+            }],
+        };
+
+        assert!(!matcher.matches(&event).unwrap());
+        assert!(matcher
+            .matches(&Event {
+                message: "doggy".to_owned(),
+                ..event
+            })
+            .unwrap());
```
evanrelf, line 70: Feels strange to me that `matches` is mutating the matcher. I feel like consuming events and mutating state should be separate from checking whether the matcher is satisfied yet.

Something like this (dumb names aside):

```rust
matcher.consume(&event).unwrap();
assert!(!matcher.is_satisfied());

matcher.consume(&Event { message: "doggy".to_owned(), ..event }).unwrap();
assert!(matcher.is_satisfied());
```

Where `consume` takes a `&mut self` and `is_satisfied` takes a `&self`.

test-harness/src/matcher/option_matcher.rs
```diff
@@ -0,0 +1,118 @@
...
+
+use crate::IntoMatcher;
+use crate::Matcher;
+
+use super::NeverMatcher;
+
+/// A matcher which may or may not contain a matcher.
+///
+/// If it does not contain a matcher, it never matches.
+pub struct OptionMatcher<M>(Option<M>);
```
evanrelf, line 11: Could you instead `impl<M: Matcher> Matcher for Option<M>` or something like that?

evanrelf (approved):
Left some thoughts, but nothing sticks out as incorrect or anything, so not blocking a merge.

test-harness/src/ghciwatch.rs
9999years: Oh, this is a horrible quirk of Rust's `async`/`await` compiling down into state machines: an `async` block like the one used here is like a separate function, so the `return` on line 337 only returns from the `async` block passed to `tokio::time::timeout`.

- The `Ok(Ok(event))` branch is reached from the `async` block return on line 337.
- The `Ok(Err(_))` branch is reached from the try (`?`) expressions on lines 335-336.
- The `Err(_)` branch is reached if the timeout expires.

test-harness/src/ghciwatch.rs
9999years: Nice catch

test-harness/src/ghciwatch.rs
evanrelf: Ah, that makes sense! I knew about the async/await state machine stuff, but I didn't realize `return` was "scoped" to an `async` block. TIL.

test-harness/src/matcher/and_matcher.rs
9999years: Hm. My thought process is that the only thing that can update whether the matcher is satisfied is feeding it an event. Splitting it up like this would also mean storing some extra state, like `FusedMatcher` does.

This is a parallel to the [`Iterator::next`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#tymethod.next) method:

> Returns [`None`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#variant.None) when iteration is finished. Individual iterator implementations may choose to resume iteration, and so calling `next()` again may or may not eventually start returning [`Some(Item)`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#variant.Some) again at some point.

To compensate for this, there's a [`FusedIterator`](https://doc.rust-lang.org/stable/std/iter/trait.FusedIterator.html) trait:

> An iterator that always continues to yield None when exhausted.
>
> Calling `next` on a fused iterator that has returned `None` once is guaranteed to return [`None`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#variant.None) again. This trait should be implemented by all iterators that behave this way because it allows optimizing [`Iterator::fuse()`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.fuse).

test-harness/src/matcher/option_matcher.rs
9999years: Not without writing my own `Display` replacement/wrapper trait at least:

```
error[E0277]: `std::option::Option<M>` doesn't implement `std::fmt::Display`
  --> test-harness/src/matcher/option_matcher.rs:13:30
   |
13 | impl<M: Matcher> Matcher for Option<M> {}
   |                              ^^^^^^^^^ `std::option::Option<M>` cannot be formatted with the default formatter
   |
   = help: the trait `std::fmt::Display` is not implemented for `std::option::Option<M>`
   = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
note: required by a bound in `Matcher`
  --> test-harness/src/matcher/mod.rs:30:20
   |
30 | pub trait Matcher: Display {
   |                    ^^^^^^^ required by this bound in `Matcher`
```

test-harness/src/ghciwatch.rs
9999years: It's kind of evil because `async` is the only block scoped this way. There's been [a proposal for `try {}` blocks that behave similarly](https://doc.rust-lang.org/beta/unstable-book/language-features/try-blocks.html) since at least 2016 but there's some issues with type inference that have kept it from getting merged.

test-harness/src/matcher/and_matcher.rs
9999years: > I'm struggling to see the connection to Rust's iterators here (mayyybe it feels kinda like a peekable iterator?)

You're correct, [`Peekable`](https://doc.rust-lang.org/stable/std/iter/struct.Peekable.html) is similar, except instead of storing if the iterator has ever returned `None`, it stores the iterator's next item. 

For some `Matcher`s (like the `BaseMatcher`), the matching is stateless and pure; if we wanted to have separate `consume` and `is_satisfied` methods, `BaseMatcher` would need to store an `is_satisfied: boolean` field to remember if it has already matched. `FusedMatcher` behaves like this, storing an extra `matched: boolean` field and returning `Ok(true)` from `Matcher::matches` unconditionally if `matched` is `true`.

Similarly, `Iterator`'s `next()` method mutates the iterator's state to (possibly) provide a next element. Because `Iterator` doesn't have an `is_finished` method, returning `None` from `next()` once doesn't mean it will never return `Some` again (e.g. for an `Iterator` reading lines from a file, while some other process writes data to the file). So `std` provides a `FusedIterator` that stores an extra bit of data: has the underlying iterator returned `None` yet? If it has, `Iterator::next` always returns `None`, and otherwise it calls the underlying method.

It's the same principle, adding an extra bit of state to make another guarantee on the API.

9999years: The underlying reason for `Peekable` is the same, too -- sometimes it's easy to check if there's a next element in advance, sometimes it's not, but it's always easy to take an unpeekable iterator, call `next` on it, and memoize the return value.

evanrelf: Ah okay, I see how it's similar to `Iterator` now, thanks for explaining.

evanrelf (approved):

PR #125, opened 2023-10-04
Fix "modules defined in multiple files" bug when modules are renamed

9999years:
This fixes a "modules defined in multiple files" bug when a module is renamed. The rename is interpreted as one module being deleted (which triggers a restart) and one module being added (which triggers an `:add`).

The bug is that it would restart `ghci`, thereby loading the new module, and then attempt to `:add` the new module anyways, which would fail.

### DIFF
--- a/src/ghci/mod.rs
+++ b/src/ghci/mod.rs
@@ -380,6 +380,9 @@ impl Ghci {
                 tracing::info!(%command, "Running after-restart command");
                 self.stdin.run_command(&mut self.stdout, command).await?;
             }
+            // Once we restart, everything is freshly loaded. We don't need to add or
+            // reload any other modules.
+            return Ok(());
         }
 
         if actions.needs_add_or_reload() {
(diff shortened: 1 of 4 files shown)


src/ghci/mod.rs
```diff
@@ -380,6 +380,9 @@ impl Ghci {
                 tracing::info!(%command, "Running after-restart command");
                 self.stdin.run_command(&mut self.stdout, command).await?;
             }
+            // Once we restart, everything is freshly loaded. We don't need to add or
+            // reload any other modules.
+            return Ok(());
```
9999years, line 385: This is the actual bug fix.

src/ghci/mod.rs
```diff
@@ -551,6 +554,11 @@ impl Ghci {
     /// Optionally returns a compilation result.
     #[instrument(skip(self), level = "debug")]
     async fn add_module(&mut self, path: &NormalPath) -> miette::Result<Option<CompilationResult>> {
+        if self.targets.contains_source_path(path.absolute())? {
+            tracing::debug!(%path, "Skipping `:add`ing already-loaded path");
+            return Ok(None);
+        }
```
9999years, line 560: This is _also_ sufficient to fix the bug. I'm adding it here in case I make this programming error in the future (calling `Ghci::add` on a file that's already part of the target set).

Gabriella439 (approved):

PR #141, opened 2023-10-16
Support `async:` commands

9999years:
This lets you prefix shell commands with `async:` to have them run asynchronously. This will let us do things like reindex hie files in the background.


parsonsmatt (approved):
Woo! This'll be a nice upgrade :)

PR #153, opened 2023-11-09
Parse eval commands when they end with EOF instead of a newline

9999years:
With an eval command at the end of the file and no newline at the end of the file, parsing fails.

This happens because [VS Code is a bad text editor](https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files) that doesn't respect [POSIX](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline), which defines a line as:

> **3.206 Line**
> 
> A sequence of zero or more non- \<newline> characters plus a terminating \<newline> character.

No terminating newline? Not a line.

Anyways, we have to work around this because people aren't exactly going to stop using VS Code.

### DIFF
--- a/src/ghci/parse/lines.rs
+++ b/src/ghci/parse/lines.rs
@@ -1,18 +1,25 @@
-use winnow::combinator::terminated;
+use winnow::ascii::line_ending;
+use winnow::ascii::not_line_ending;
+use winnow::combinator::alt;
+use winnow::combinator::eof;
+use winnow::error::ContextError;
+use winnow::error::ErrMode;
 use winnow::stream::AsChar;
+use winnow::stream::Compare;
 use winnow::stream::FindSlice;
+use winnow::stream::SliceLen;
 use winnow::stream::Stream;
 use winnow::stream::StreamIsPartial;
-use winnow::token::take_until0;
 use winnow::PResult;
 use winnow::Parser;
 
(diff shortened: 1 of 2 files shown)


src/ghci/parse/lines.rs
```diff
@@ -1,18 +1,25 @@
...
 use winnow::stream::StreamIsPartial;
-use winnow::token::take_until0;
 use winnow::PResult;
 use winnow::Parser;
 
 /// Parse the rest of a line, including the newline character.
 pub fn rest_of_line<I>(input: &mut I) -> PResult<<I as Stream>::Slice>
 where
-    I: Stream + StreamIsPartial + for<'i> FindSlice<&'i str>,
+    I: Stream + StreamIsPartial + for<'i> FindSlice<&'i str> + for<'i> Compare<&'i str>,
```
parsonsmatt, line 19: What does the `for` keyword do here?

parsonsmatt (approved):
Great job! One question for my own curiosity

src/ghci/parse/lines.rs
9999years: It's basically a `forall`. Apparently they're called [higher ranked trait bounds](https://doc.rust-lang.org/nomicon/hrtb.html).

PR #156, opened 2023-11-20
Extract `ghci` restart logic

9999years:
NFCI


atruslow (approved):
This looks good to me 👍

PR #158, opened 2023-11-21
Interrupt reloads when new files change

9999years:
This will give a snappier UX in general.

Note that now, many hooks can be canceled or mismatched (e.g. the before-reload hooks can run but the after-reload hooks get canceled). I'm not sure if this will cause issues with Honeycomb (e.g. spans that open but don't close).

### DIFF
--- a/src/command_ext.rs
+++ b/src/command_ext.rs
@@ -1,13 +1,5 @@
 use std::process::Command as StdCommand;
 
-use command_group::AsyncCommandGroup;
-use command_group::AsyncGroupChild;
-use miette::Context;
-use miette::IntoDiagnostic;
-use nix::sys::signal::pthread_sigmask;
-use nix::sys::signal::SigSet;
-use nix::sys::signal::SigmaskHow;
-use nix::sys::signal::Signal;
 use tokio::process::Command;
 
 /// Extension trait for commands.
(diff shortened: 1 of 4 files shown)


src/command_ext.rs
```diff
@@ -35,42 +27,3 @@ impl CommandExt for StdCommand {
...
-        })
-    }
-}
-
-fn spawn_without_inheriting_sigint<T>(
-    spawn: impl FnOnce() -> miette::Result<T>,
-) -> miette::Result<T> {
-    // See: https://github.com/rust-lang/rust/pull/100737#issuecomment-1445257548
-    let mut old_signal_mask = SigSet::empty();
-    pthread_sigmask(
```
9999years, line 64: It turns out that this doesn't just stop `ghci` from _inheriting_ `SIGINT` from `ghciwatch`, it also stops `ghci` from receiving `SIGINT` whatsoever.

This is a little messier, I guess — when users press Ctrl-C the underlying `ghci` session will see that as well — but then `ghciwatch` will shut down, so it doesn't matter much in the end.

src/ghci/manager.rs
```diff
@@ -51,29 +70,62 @@ pub async fn run_ghci(
...
+            reload_sender,
+        )));
         tokio::select! {
             _ = handle.on_shutdown_requested() => {
                 // Cancel any in-progress reloads. This releases the lock so we don't block here.
                 task.abort();
                 ghci.lock().await.stop().await.wrap_err("Failed to quit ghci")?;
                 break;
             }
+            Some(new_event) = receiver.recv() => {
```
Gabriella439, line 112: Might be worth explicitly matching on the `Reload` constructor of `GhciEvent` so that if you later add more constructors this will become a non-exhaustive pattern match.

Gabriella439 (approved):

PR #159, opened 2023-11-21
Allow customizing load-bearing sleeps

9999years:
The test suite features a load-bearing sleep of 1 second before writing. This makes the filesystem events get detected more reliably, for whatever reason.

This change allows disabling the load-bearing sleeps; this is typically needed to make several writes that all get detected together by `ghciwatch`.

### DIFF
--- a/test-harness/src/ghciwatch.rs
+++ b/test-harness/src/ghciwatch.rs
@@ -19,6 +19,7 @@ use crate::BaseMatcher;
 use crate::Checkpoint;
 use crate::CheckpointIndex;
 use crate::Event;
+use crate::Fs;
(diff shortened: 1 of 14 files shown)


test-harness/src/ghciwatch.rs
```diff
@@ -477,16 +482,26 @@ impl GhciWatch {
     pub fn pid(&self) -> u32 {
         self.pid
     }
+
+    /// Get the filesystem helpers.
+    pub fn fs(&self) -> &Fs {
```
Gabriella439, line 487: maybe make this a public field so you don't need the method

Gabriella439 (approved):

PR #161, opened 2023-11-22
Add more lifecycle shell hooks

9999years:
Forgot to add these. Probably worth making a better system with less of a combinatorial explosion to deal with.


Gabriella439 (approved):
This looks like it would conflict with https://github.com/MercuryTechnologies/ghciwatch/pull/161 but I assume you'll sort that out

PR #162, opened 2023-11-22
Store lifecycle hooks dynamically

9999years:
Instead of having a field for each event (`before_restart_ghci`, `test_ghci`, `after_startup_shell`, etc.), we generate and store the arguments dynamically. This simplifies the running of hooks and will hopefully make it easier to extend the lifecycle hooks in the future.

### DIFF
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -1,0 +1,393 @@
+//! Lifecycle hooks.
+
+use std::fmt::Display;
+use std::process::ExitStatus;
+use std::str::FromStr;
+
+use clap::builder::ValueParserFactory;
+use clap::Arg;
+use clap::ArgAction;
+use clap::Args;
+use clap::FromArgMatches;
+use strum::EnumMessage;
+use strum::EnumProperty;
+use strum::IntoEnumIterator;
+use tokio::task::JoinHandle;
+
+use crate::ghci::GhciCommand;
+use crate::maybe_async_command::MaybeAsyncCommand;
+
+/// A lifecycle event that triggers hooks.
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    strum::EnumVariantNames,
+    strum::EnumIter,
+    strum::EnumMessage,
+    strum::EnumProperty,
+    strum::Display,
+    strum::AsRefStr,
+)]
+#[strum(serialize_all = "kebab-case")]
+pub enum LifecycleEvent {
+    /// When tests are run (after startup, after reloads).
+    #[strum(
+        message = "
+        Tests are run after startup and after reloads.
+        ",
+        props(help_name = "tests")
+    )]
+    Test,
+    /// When a `ghci` session is started (at `ghciwatch` startup and after restarts).
+    #[strum(message = "
+        Startup hooks run when `ghci` is started (at `ghciwatch` startup and after `ghci` restarts).
+    ")]
+    Startup,
+    /// When a module is changed or added.
+    #[strum(message = "
+        Reload hooks are run when modules are changed on disk.
+    ")]
+    Reload,
+    /// When a `ghci` session is restarted (when a module is removed or renamed).
+    #[strum(message = "
+        `ghci` is restarted when modules are removed or renamed.
+        See: https://gitlab.haskell.org/ghc/ghc/-/issues/11596
+    ")]
+    Restart,
+}
+
+impl LifecycleEvent {
+    fn supported_when(&self) -> Vec<When> {
+        match self {
+            LifecycleEvent::Test => vec![When::During],
+            LifecycleEvent::Startup | LifecycleEvent::Reload | LifecycleEvent::Restart => {
+                vec![When::Before, When::After]
+            }
+        }
+    }
+
+    fn supported_kind(&self, when: When) -> Vec<CommandKind> {
+        debug_assert!(self.supported_when().contains(&when));
+        match self {
+            LifecycleEvent::Reload | LifecycleEvent::Restart | LifecycleEvent::Test => {
+                vec![CommandKind::Ghci, CommandKind::Shell]
+            }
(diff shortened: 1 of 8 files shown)


src/hooks.rs
```diff
@@ -0,0 +1,396 @@
...
+
+impl LifecycleEvent {
+    fn supported_when(&self) -> Vec<When> {
+        match self {
+            LifecycleEvent::Test => vec![When::During],
+            LifecycleEvent::Startup | LifecycleEvent::Reload | LifecycleEvent::Restart => {
+                vec![When::Before, When::After]
+            }
+        }
+    }
```
Gabriella439, line 72: I think rather than check at runtime that we don't have an invalid combination of `LifecycleEvent` and `When` we should enforce that statically by combining them into a single data structure using something like this:

```rust
pub enum Bounded {
    Before,
    After,
}

pub enum LifecycleEvent {
    Test,
    StartUp{ when: When },
    Reload{ when: When },
    Restart{ when: When },
}
```

9999years: I did consider this, but it gets kind of clumsy; for one, I'd have to add an explicit list of each possible event:

```rust
fn events() -> Vec<LifecycleEvent> {
  vec![
    LifecycleEvent::Test,
    LifecycleEvent::StartUp(When::Before),
    LifecycleEvent::StartUp(When::After),
    // ...
  ]
}
```

With that approach, when I add a new lifecycle event, nothing will tell me to add the new event to the list, and I also have to make sure to include every possible variant.

Gabriella439: Wouldn't that be an issue even for the existing approach?

evanrelf: `supported_when` (the function you highlighted) matches on a value of type `LifecycleEvent`. The pattern match would become non-exhaustive.

src/hooks.rs
```diff
@@ -0,0 +1,396 @@
...
+        handles: &mut Vec<JoinHandle<miette::Result<ExitStatus>>>,
+    ) -> miette::Result<()> {
+        for hook in self.select(event, when) {
+            match &hook.command {
+                Command::Ghci(_) => {}
+                Command::Shell(command) => {
+                    tracing::info!(%command, "Running {hook} command");
+                    command.run_on(handles).await?;
+                }
+            }
```
evanrelf, line 320:
```suggestion
            if let Command::Shell(command) = &hook.command {
                tracing::info!(%command, "Running {hook} command");
                command.run_on(handles).await?;
            }
```

src/hooks.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// When a `ghci` session is restarted (when a module is removed or renamed).
+    #[strum(message = "
+        `ghci` is restarted when modules are removed or renamed.
+        See: https://gitlab.haskell.org/ghc/ghc/-/issues/11596
+    ")]
+    Restart,
+}
+
+impl LifecycleEvent {
+    fn supported_when(&self) -> Vec<When> {
```
evanrelf, line 65:
```suggestion
    fn supported_when(&self) -> HashSet<When> {
```

src/hooks.rs
```diff
@@ -0,0 +1,396 @@
...
+        match self {
+            LifecycleEvent::Test => vec![When::During],
+            LifecycleEvent::Startup | LifecycleEvent::Reload | LifecycleEvent::Restart => {
+                vec![When::Before, When::After]
+            }
+        }
+    }
+
+    fn supported_kind(&self, when: When) -> Vec<CommandKind> {
```
evanrelf, line 74:
```suggestion
    fn supported_kind(&self, when: When) -> HashSet<CommandKind> {
```

src/hooks.rs
```diff
@@ -0,0 +1,396 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.when {
+            When::During => write!(f, "{}", self.event.as_ref()),
+            _ => write!(f, "{}-{}", self.when, self.event),
+        }
+    }
+}
+
+impl<C> Hook<C> {
+    fn with_command<C2>(&self, command: C2) -> Hook<C2> {
```
evanrelf, line 192: Is the `C2` necessary? `C` isn't constrained in any way, so I think it would always unify with whatever the type of `command` is.

evanrelf (approved):

evanrelf:
This is 100% overkill but

You could replace `HashSet<T>` (where `T` is `When` or `CommandKind`) with `FlagSet<T>` from the [flagset](https://docs.rs/flagset) crate (nicer alternative to `bitflags` if you're familiar with that).

src/hooks.rs
9999years: And I have a proc macro that generates a function which iterates over all the enum variants, which I can only use if they don't contain data.

Gabriella439: it is true that in that one case you might lose exhaustiveness, but on the balance I feel like you gain better safety/exhaustiveness by making invalid states unrepresentable

src/hooks.rs
9999years: Yeah, because this lets you change the generic type (i.e. it lets you go from a `Hook<CommandKind>` to a `Hook<Command>`). That's also why I can't use the `..self` record update syntax here — the types need to match exactly.

src/hooks.rs
9999years: Consider that the impetus of this system is that I [actually already forgot to fill in new entries in the old lifecycle hooks list](https://github.com/MercuryTechnologies/ghciwatch/pull/161)! This system prevents that. If there's a way to implement this data type that makes invalid states unrepresentable _and_ doesn't require me to explicitly list all the valid states, I'll gladly implement it.

Gabriella439: Can you clarify which list you're referring to?  I understand what you're asking for in the abstract but couldn't place where in each pull request this corresponds to

Gabriella439 (approved):

PR #163, opened 2023-11-27
Cleanup Nix code

9999years:
Gives us actual packages, makes the definitions all a bit nicer.


nix/packagesFromDirectory.nix
```diff
@@ -0,0 +1,65 @@
+{
```
Gabriella439, line 1: I think there's still the old version of this function in `nix/overlays/lib.nix`, which you can probably delete now

flake.nix
```diff
@@ -33,108 +31,99 @@
...
-            ];
-          };
-
-        pkgs = makePkgs {};
+    advisory-db,
+    flake-compat,
+  }: let
+    eachSystem = nixpkgs.lib.genAttrs (import systems);
+
+    makeNixpkgs = {
```
Gabriella439, line 45: Usually for Nix code I reserve the term "nixpkgs" to refer to the Nixpkgs source code and use the term "pkgs" or "packages" to refer to the actual attribute set of packages instantiated from the Nixpkgs source code.  So I'd suggest calling this utility `makePkgs` instead of `makeNixpkgs` (also because it creates the `pkgs` variable, so the name would more closely match the result it is used to create).

Gabriella439 (commented):
looks great!  just two things

nix/packagesFromDirectory.nix
9999years: That's in `mercury-web-backend`!

Gabriella439: Oh, whoops!  I didn't notice this was a different repo 😅.  Never mind me, then.

Gabriella439 (approved):

PR #170, opened 2023-12-01
Don't print "All good!" when compilation is not "All good!"

9999years:
Turned out I wasn't ever setting the "compilation failed" variable where I expected it.

Also I had a bug with the `BaseMatcher`, so I fixed it to be stricter about spans.

### DIFF
--- a/tests/all_good.rs
+++ b/tests/all_good.rs
@@ -1,0 +1,45 @@
+use test_harness::test;
+use test_harness::BaseMatcher;
+use test_harness::GhciWatchBuilder;
+use test_harness::Matcher;
+
+/// Test that `ghciwatch` can detect when compilation fails.
+///
+/// Regression test for DUX-1649.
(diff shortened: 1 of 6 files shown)


parsonsmatt (approved):

tests/all_good.rs
```diff
@@ -0,0 +1,45 @@
+use test_harness::test;
+use test_harness::BaseMatcher;
+use test_harness::GhciWatchBuilder;
+use test_harness::Matcher;
+
+/// Test that `ghciwatch` can detect when compilation fails.
+///
+/// Regression test for DUX-1649.
```
Gabriella439, line 8: Perhaps create a matching GitHub issue that can be referenced here since this is an open source project (since outside contributors won't have access to our internal Linear).

Gabriella439 (approved):

PR #174, opened 2023-12-05
Make the compilation failed message visible

9999years:
Make the "compilation failed" message error-level rather than debug-level. The `ghci` error output does generally show when compilation fails, but without a visible prompt it's not super clear when exactly compilation finished. This is a follow-up to #167 + #170, which added an "All good!" message.


parsonsmatt (approved):
awesome! That's a great change

PR #175, opened 2023-12-05
Make GHCi output destination configurable

evanrelf:
Context:

I'm working on adding a TUI mode, and I'd like GHCi's output to go to an in-memory buffer instead of to `stdout` and `stderr`, so I can render it as I see fit.

This change will allow me to drop in a pair of `tokio::io::DuplexStream`s in place of `tokio::io::{Stdout,Stderr}`.

Worth mentioning that this does _not_ affect output from `tracing`. It only controls the `ghci` process' output.


src/ghci/stderr.rs
```diff
@@ -87,11 +97,18 @@ impl GhciStderr {
...
+        line.push('\n');
         self.buffer.push_str(&line);
-        self.buffer.push('\n');
-        eprintln!("{line}");
+        self.writer
+            .write_all(line.as_bytes())
+            .await
+            .into_diagnostic()?;
+        Ok(())
     }
```
evanrelf, line 112: I removed `self.buffer.push('\n');` in favor of `line.push('\n');`. `write_all` wouldn't add a trailing newline like `eprintln` did, so I need the newline in both places.

src/ghci/write.rs
```diff
@@ -0,0 +1,33 @@
...
+impl<W> IntoGhciWrite for W
+where
+    W: AsyncWrite + TokioAsyncWriteCompatExt + Unpin + Send + Sized + Debug + 'static,
+{
+    fn into_ghci_write(self) -> Box<dyn GhciWrite> {
+        Box::new(FuturesAsyncWriteCompatExt::compat_write(Arc::new(
+            Mutex::new(TokioAsyncWriteCompatExt::compat_write(self)),
+        )))
+    }
+}
```
evanrelf, line 33: This is the ugliest part of the PR. Interop between `tokio` and `futures` is annoying, and `Box` + `Arc` + `Mutex` isn't the most efficient.

Cargo.toml
```diff
@@ -34,10 +34,12 @@ categories = ["command-line-utilities", "development-tools"]
 
 [dependencies]
 aho-corasick = "1.0.2"
+async-dup = "1.2.4"
 backoff = { version = "0.4.0", default-features = false }
 camino = "1.1.4"
 clap = { version = "4.3.2", features = ["derive", "wrap_help", "env", "string"] }
 command-group = { version = "2.1.0", features = ["tokio", "with-tokio"] }
+dyn-clone = "1.0.16"
```
evanrelf, line 42: I don't know how `dyn-clone` works, but it's `dtolnay` so I trust it lol.

Cargo.toml
```diff
@@ -57,6 +59,7 @@ strip-ansi-escapes = "0.2.0"
 tap = "1.0.1"
 textwrap = { version = "0.16.0", features = ["terminal_size"] }
 tokio = { version = "1.28.2", features = ["full", "tracing"] }
+tokio-util = { version = "0.7.10", features = ["compat"] }
```
evanrelf, line 62: The `smol` ecosystem has their own `async-compat` crate, but their `Compat` newtype wrapper thing doesn't implement `Clone` like `tokio-util`'s does, so I use `tokio-util` here.

Cargo.toml
```diff
@@ -34,10 +34,12 @@ categories = ["command-line-utilities", "development-tools"]
 
 [dependencies]
 aho-corasick = "1.0.2"
+async-dup = "1.2.4"
```
evanrelf, line 37: From the `smol` ecosystem. Provides `Arc` and `Mutex` variants that implement `AsyncWrite`.

src/ghci/mod.rs
```diff
@@ -120,6 +128,8 @@ impl GhciOpts {
             hooks: opts.hooks.clone(),
             restart_globs: opts.watch.restart_globs()?,
             reload_globs: opts.watch.reload_globs()?,
+            stdout_writer: tokio::io::stdout().into_ghci_write(),
+            stderr_writer: tokio::io::stderr().into_ghci_write(),
```
9999years, line 132: So your thinking is here we could have some `--tui` flag or similar and use it to set some alternate writers here?

src/ghci/stderr.rs
```diff
@@ -19,17 +22,21 @@ pub enum StderrEvent {
     GetBuffer { sender: oneshot::Sender<String> },
 }
 
-pub struct GhciStderr {
+pub struct GhciStderr<W> {
```
9999years, line 25: If `W` is always `Box<dyn GhciWrite>`, we might just want to make that the type.

src/ghci/write.rs
```diff
@@ -0,0 +1,33 @@
...
+use std::fmt::Debug;
+use tokio::io::AsyncWrite;
+use tokio_util::compat::FuturesAsyncWriteCompatExt;
+use tokio_util::compat::TokioAsyncWriteCompatExt;
+
+/// An [`AsyncWrite`]r usable in `GhciOpts`. In particular, it must implement [`Clone`], [`Send`],
+/// and [`Sync`].
+///
+/// Use [`IntoGhciWrite`] to convert a qualifying [`AsyncWrite`]r into a [`GhciWrite`]r.
+pub trait GhciWrite: AsyncWrite + DynClone + Debug + Send + Sync + Unpin {}
```
9999years, line 13: Good lord.

src/ghci/mod.rs
```diff
@@ -98,6 +102,10 @@ pub struct GhciOpts {
     pub restart_globs: GlobMatcher,
     /// Reload the `ghci` session when paths matching these globs are changed.
     pub reload_globs: GlobMatcher,
+    /// Where to write what `ghci` emits to `stdout`. Inherits parent's `stdout` by default.
+    pub stdout_writer: Box<dyn GhciWrite>,
+    /// Where to write what `ghci` emits to `stderr`. Inherits parent's `stderr` by default.
+    pub stderr_writer: Box<dyn GhciWrite>,
```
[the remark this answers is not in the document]
9999years: Maybe instead of type parameters or `Box<dyn>`, we could have an enum with all the types we expect:

```rust
enum GhciStdout {
  Stdout(...),
  Tui(...),
}
```

And then `impl` `AsyncWrite` and whatnot on that?

I'd really love to avoid `Box<dyn Arc<Mutex<T>>>`, but I suspect in practice the overhead won't be too bad if we can't find something nicer.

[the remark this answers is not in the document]
evanrelf: > Maybe instead of type parameters or Box<dyn>, we could have an enum with all the types we expect

Good idea! That'd remove the `Box` in all cases, and the `Arc` + `Mutex` in the `Stdout`/`Stderr` cases (`clone` would just create another via `tokio::io::std{out,err}()`).

> I suspect in practice the overhead won't be too bad

I agree. I think it's more objectionable in theory than in practice. I think the terminal and the user's brain have lower throughput than this lol.

src/ghci/stderr.rs
evanrelf: Done in https://github.com/MercuryTechnologies/ghciwatch/pull/175/commits/388f83624873866f142dcb09d63be112daac9d58 and https://github.com/MercuryTechnologies/ghciwatch/pull/175/commits/8d0e44680084b74e6b01ea8a8a21ec8cc29fe081

src/ghci/mod.rs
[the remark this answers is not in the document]
evanrelf: I implemented this in https://github.com/MercuryTechnologies/ghciwatch/pull/175/commits/145f5ec71e0ebebd850f544c1816c1dd528ef87b.

I think the idea you mentioned in Slack - keeping `AsyncWrite`rs out of `GhciOpts`, opting to create them during startup or restart and send them over a channel - is a good one, but I'd rather explore that in a future PR if that's okay.

Cargo.toml
evanrelf: Removed in https://github.com/MercuryTechnologies/ghciwatch/pull/175/commits/508bcf98521987a80928ab2f0570b2fcbdc71a81, it's no longer needed after https://github.com/MercuryTechnologies/ghciwatch/pull/175/commits/145f5ec71e0ebebd850f544c1816c1dd528ef87b

9999years (approved):

PR #183, opened 2023-12-07
Make GhciWriter public

evanrelf:
I need to use it from `main` in the bin target.


9999years (approved):

9999years:
I think this is fine, though I'd prefer to keep as much logic out of `main.rs` as possible.

PR #185, opened 2023-12-09
Add primitive TUI mode

evanrelf:
Shows the same GHCi output and logging as the current print-to-`stdout` behavior, but renders it as a TUI. This mode is off by default, enabled by passing an undocumented `--tui` flag.


Cargo.toml
```diff
@@ -52,15 +55,18 @@ once_cell = "1.18.0"
...
 strip-ansi-escapes = "0.2.0"
 tap = "1.0.1"
+tempfile = "3.8.1"
 textwrap = { version = "0.16.0", features = ["terminal_size"] }
 tokio = { version = "1.28.2", features = ["full", "tracing"] }
+tokio-stream = { version = "0.1.14", default-features = false }
 tokio-util = { version = "0.7.10", features = ["compat"] }
 tracing = "0.1.37"
-tracing-human-layer = "0.1.0"
+tracing-human-layer = "0.1.2"
```
evanrelf, line 69: Updated to get `HumanLayer::with_output_stream`

Cargo.toml
```diff
@@ -34,12 +34,15 @@ categories = ["command-line-utilities", "development-tools"]
 
 [dependencies]
 aho-corasick = "1.0.2"
+ansi-to-tui = "3.1.0"
```
evanrelf, line 37: Converts a byte slice into `ratatui`'s `Text` type, which represents formatted text. Without this, the escape codes are printed to the screen as regular characters.

Cargo.toml
```diff
@@ -34,12 +34,15 @@ categories = ["command-line-utilities", "development-tools"]
...
 async-dup = "1.2.4"
 backoff = { version = "0.4.0", default-features = false }
 camino = "1.1.4"
 clap = { version = "4.3.2", features = ["derive", "wrap_help", "env", "string"] }
 command-group = { version = "2.1.0", features = ["tokio", "with-tokio"] }
+crossterm = { version = "0.27.0", features = ["event-stream"] }
 enum-iterator = "1.4.1"
+gag = "1.0.0"
```
evanrelf, line 45: I'm using `gag` to capture the `ghciwatch` process' `stderr` in a buffer so I can render it with the TUI.

I would prefer to specify a `Write`r at the `tracing-subscriber` level, instead of hackily capturing the output after it's emitted, but `tracing-human-layer` only supports emitting 
to `stdout` or `stderr` today.

Cargo.toml
```diff
@@ -52,15 +55,18 @@ once_cell = "1.18.0"
 owo-colors = { version = "3.5.0", features = ["supports-colors"] }
 path-absolutize = "3.1.1"
 pathdiff = { version = "0.2.1", features = ["camino"] }
+ratatui = "0.24.0"
 regex = { version = "1.9.3", default-features = false, features = ["perf", "std"] }
 shell-words = "1.1.0"
 strip-ansi-escapes = "0.2.0"
 tap = "1.0.1"
+tempfile = "3.8.1"
```
evanrelf, line 63: `gag` uses a temporary file to redirect `stdout` and `stderr`. Apparently it's guaranteed to be in memory with modern Linux?

I need `tempfile` for the `AsyncBufferRedirect` thing.

src/async_buffer_redirect.rs
```diff
@@ -0,0 +1,70 @@
...
+use tokio::fs::File;
+use tokio::io::AsyncRead;
+use tokio::io::ReadBuf;
+
+/// Buffer output in an in-memory buffer. Async version of [`gag::BufferRedirect`].
+pub struct AsyncBufferRedirect {
+    #[allow(dead_code)]
+    redir: Redirect<StdFile>,
+    outer: File,
+}
```
evanrelf, line 17: Exactly the same as `gag`'s `BufferRedirect` type, except it holds a `tokio::fs::File` instead of a `std::fs::File` so I can use `AsyncRead` instead of hacking together some `tokio::task::spawn_blocking` + `Read` thing.

src/async_buffer_redirect.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// An in-memory read-only buffer into which [`AsyncBufferRedirect`] buffers output. Async version
+/// of [`gag::Buffer`].
+pub struct AsyncBuffer(File);
+
+impl AsyncRead for AsyncBuffer {
+    #[inline(always)]
+    fn poll_read(
```
evanrelf, line 25: I'm copying the inline attributes from the original `BufferRedirect` implementation.

src/tracing/mod.rs
```diff
@@ -32,6 +35,11 @@ impl<'opts> TracingOpts<'opts> {
             filter_directives: &opts.logging.tracing_filter,
             trace_spans: &opts.logging.trace_spans,
             json_log_path: opts.logging.log_json.as_deref(),
+            output_stream: if opts.tui {
+                OutputStream::Stderr
+            } else {
+                OutputStream::Stdout
+            },
```
evanrelf, line 42: `tracing-human-layer` uses `OutputStream::Stdout` by default, so this is behavior preserving when you're not in TUI mode.

src/tui/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    Ok(())
+}
+
+const SCROLL_AMOUNT: usize = 1;
+
+fn handle_event(tui: &mut Tui, event: Event) -> miette::Result<()> {
+    match event {
+        // TODO(evan): Scrolling is excruciatingly slow
+        Event::Mouse(mouse) if mouse.kind == MouseEventKind::ScrollUp => {
```
evanrelf, line 124: I think slow scrolling (and terminal event handling in general, including `Ctrl-C` and window resizing) is caused by slow, effectively blocking reads (of GHCi output and `tracing` logs on `stderr`).

src/tui/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    Paragraph::new(text)
+        .wrap(Wrap::default())
+        .scroll((scroll_offset, 0))
+        .render(area, buffer);
+
+    Ok(())
+}
+
+const SCROLL_AMOUNT: usize = 1;
```
evanrelf, line 119: My terminal scrolls by one line at a time, but Vim and other text editors typically scroll by 3 lines at a time. Not sure which is best to pick, or whether this should be configurable. Not important right now though.

src/tui/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    Ok(())
+}
+
+fn render(tui: &Tui, area: Rect, buffer: &mut Buffer) -> miette::Result<()> {
+    if area.width == 0 || area.height == 0 {
+        return Ok(());
+    }
+
+    let text = tui.scrollback.into_text().into_diagnostic()?;
+
```
evanrelf, line 108: I would rather rendering be infallible, but the conversion to UTF-8 is done as part of the `ratatui` `Text` building process, so for now it lives here.

I haven't had issues with non-UTF-8 text yet, but I imagine that's something I will need to handle (I think other parts of `ghciwatch` have to handle this?).

src/tui/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    let _ = shutdown.request_shutdown();
+
+    Ok(())
+}
+
+fn render(tui: &Tui, area: Rect, buffer: &mut Buffer) -> miette::Result<()> {
+    if area.width == 0 || area.height == 0 {
+        return Ok(());
+    }
```
evanrelf, line 105: Prevents unnecessary work and potential panics (like your rendering code assumes something, not that `ratatui` did anything wrong) when there's nothing to render.

src/tui/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+                let event = output
+                    .ok_or_else(|| miette!("No more crossterm events"))?
+                    .into_diagnostic()
+                    .wrap_err("Failed to get next crossterm event")?;
+                handle_event(&mut tui, event)?;
+            }
+        }
+    }
+
+    let _ = shutdown.request_shutdown();
```
evanrelf, line 97: Ignoring any errors because that's what's done everywhere else with `request_shutdown`, and I think a possible error is "somebody else already requested shutdown"?

src/tui/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+        let mut render_result = Ok(());
+        terminal
+            .draw(|frame| {
+                let area = frame.size();
+                let buffer = frame.buffer_mut();
+                render_result = render(&tui, area, buffer);
+            })
+            .into_diagnostic()
+            .wrap_err("Failed to draw to terminal")?;
+        render_result?;
```
evanrelf, line 60: Unfortunate you can't use the try operator inside a closure like this.

src/tui/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+                }
+            }
+
+            output = tracing_reader.read(&mut tracing_buffer) => {
+                output
+                    .into_diagnostic()
+                    .wrap_err("Failed to read bytes from tracing into TUI buffer")?;
+                tui.scrollback.extend(tracing_buffer);
+                tracing_buffer = [0; 1024];
+            }
```
evanrelf, line 85: Filling and zeroing separate buffers because I can't hand out multiple exclusive references to `tui.scrollback`. There's probably a better (more efficient and/or idiomatic) way to do this?

src/tui/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+use tokio::io::AsyncReadExt;
+use tokio::io::DuplexStream;
+use tokio_stream::StreamExt;
+
+#[derive(Default)]
+struct Tui {
+    quit: bool,
+    scrollback: Vec<u8>,
+    // TODO(evan): Follow output when scrolled to bottom
+    scroll_offset: usize,
```
evanrelf, line 29: When the user is scrolled to the bottom of the scrollback (the default if they haven't scrolled at all), new output should push old output up and off the screen. Like a "follow logs" option in a CI service.

src/tui/terminal.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn drop(&mut self) {
+        let result = exit().wrap_err("Failed to exit terminal during drop");
+        // Ignore the `Result` if we're already panicking; aborting is undesirable
+        if !std::thread::panicking() {
+            result.unwrap();
+        }
+    }
+}
+
+static INSIDE: AtomicBool = AtomicBool::new(false);
```
evanrelf, line 43: Tracking whether we're "inside" or not (raw mode enabled, cursor hidden, events being captured, etc etc) so we can skip double entering or double exiting, in particular the implicit exits with `Drop` and panics.

This is preferable to removing the panic hook we installed, because we can't be sure an additional panic hook "layer" hasn't been installed (e.g. the `miette` panic hook in `main`). Safer to turn ours into a no-op post-exit and leave the panic hook undisturbed.

TODO: Skip double entering

src/tui/terminal.rs
```diff
@@ -0,0 +1,115 @@
...
+        cursor::Hide,
+        event::EnableMouseCapture,
+        event::EnableFocusChange,
+        event::EnableBracketedPaste,
+        event::PushKeyboardEnhancementFlags(
+            KEF::DISAMBIGUATE_ESCAPE_CODES
+                | KEF::REPORT_EVENT_TYPES
+                | KEF::REPORT_ALL_KEYS_AS_ESCAPE_CODES
+        ),
+    )
```
evanrelf, line 68: I could remove some of these, like focus change and bracketed paste events - but they don't make a big difference.

The spammy one that does make a dent is mouse events: did you know TUI apps can track mouse movement generally, like hover? So just moving the mouse over a TUI sends dozens of events. But we need mouse events for scrolling.

src/tui/mod.rs
evanrelf, line 85: Hmm, I'm skeptical of this... I'm not experienced with `select!` or async cancellation.

I know `AsyncReadExt::read` is cancel safe, but I don't know about the other select arms.

Even if they're all cancel safe, maybe the horrid performance has something to do with creating a new future on every loop or something.

Like maybe instead of something like this:

```rust
let foo: tokio::sync::mpsc::Receiver<usize> = /* ... */;
let bar: tokio::sync::mpsc::Receiver<usize> = /* ... */;
loop {
  select! {
    Some(msg) = foo.recv() => println!("foo received: {msg}"),
    Some(msg) = bar.recv() => println!("bar received: {msg}"),
    else => break,
  }
}
```

...I should have something like this:

```rust
let foo: tokio::sync::mpsc::Receiver<usize> = /* ... */;
let mut foo_fut = foo.recv();
let bar: tokio::sync::mpsc::Receiver<usize> = /* ... */;
let mut bar_fut = bar.recv();
loop {
  select! {
    Some(msg) = foo => {
      println!("foo received: {msg}");
      foo_fut = foo.recv();
    }
    Some(msg) = bar => {
      println!("bar received: {msg}");
      bar_fut = bar.recv();
    }
    else => break,
  }
}
```

...perhaps? Then I'm polling the same future repeatedly, rather than creating a new future on each loop, even if it's semantically the same?

I dunno. Second one's kinda awful lol, I hope that boilerplate isn't required.

evanrelf:
Ugh, no, the performance is abysmal because of the stupid `gag` thing. It's going through a temp file. Of course that's the problem, duh.

src/tui/mod.rs
evanrelf: I think I was wrong about the "re-creating the futures on each loop" concern.

But I am still skeptical of this `select!`, now because I think I should be `tokio::spawn`ing these futures outside of the `loop` (or at least outside of the `select!`) so that the reads can occur in parallel.

Currently, I think they're just happening concurrently on the same thread, because they're obscured from `tokio`'s view; it only sees a single `run_tui` future (or one of `run_tui`'s ancestors). I need to spawn them as new tasks for the executor to be able to schedule them on separate threads.

9999years (approved):
Let's ship it! Still needs tests (see #202 for my plan) and I have a couple fixups / extra features in #200, but this is good to merge so we can start experimenting.

PR #186, opened 2023-12-13
CLI flag for disabling interrupts

parsonsmatt:
This PR adds a cli flag that disables the interrupt behavior for reloads. This is my first contribution to a Rust project so I'm sure it'll take some polishing :)


src/cli.rs
```diff
@@ -49,6 +49,11 @@ pub struct Opts {
     /// Options to modify logging and error-handling behavior.
     #[command(flatten)]
     pub logging: LoggingOpts,
+
+    /// By default, ghciwatch will interrupt reloads if a file changes. If you want ghciwatch to
+    /// avoid interrupting reloads, set this flag.
+    #[arg(long = "no-interrupt-reloads")]
+    pub no_interrupt_reloads: bool,
```
parsonsmatt, line 56: I'm not super happy with a `--no-interrupt-reloads` flag - feels kind of awkward. But `--interrupt-reloads` is the default behavior.

We could switch the default back to `no-interrupt`, and then this flag can be `--interrupt-reloads`.

Made a slack convo to ask about this: 
https://mercurytechnologies.slack.com/archives/CUJ89R895/p1702483630925389

src/ghci/manager.rs
```diff
@@ -56,6 +56,7 @@ pub async fn run_ghci(
     // This function is pretty tricky! We need to handle shutdowns at each stage, and the process
     // is a little different each time, so the `select!`s can't be consolidated.
 
+    let no_interrupt_reloads = opts.no_interrupt_reloads;
```
parsonsmatt, line 59: This definitely feels awkward. I tried passing the `opts` directly in to the relevant function, but the `opts` are moved into `Ghci::new` call just below. I can get the `bool` out here no problem, though, so I do that instead.

src/ghci/manager.rs
```diff
@@ -113,19 +114,21 @@ pub async fn run_ghci(
...
-                    event.merge(new_event);
-                    maybe_event = Some(event);
-
-                    // Cancel the in-progress reload. This releases the `ghci` lock to prevent a deadlock.
-                    task.abort();
-
-                    // Send a SIGINT to interrupt the reload.
-                    // NB: This may take a couple seconds to register.
-                    ghci.lock().await.send_sigint().await?;
+                if !no_interrupt_reloads {
```
parsonsmatt, line 117: here's our behavior change: if we set this flag, then we do not go into that block

src/cli.rs
9999years: There's an open issue for automatic negation flags (i.e. adding `--interrupt-reloads` and `--no-interrupt-reloads` at once) here:

https://github.com/clap-rs/clap/issues/815

src/ghci/manager.rs
9999years: Yeah I think this is fine.

9999years (approved):

PR #192, opened 2024-02-01
Write `--errors` log before running shell hooks

9999years:
This lets users observe error log output in lifecycle hooks. Useful for some editor integrations, e.g.:

```
vim --servername $VIM_SERVERNAME --remote-send "<C-W>:cgetfile ghcid.txt<CR>"
```

Also adds a couple features to the test harness:
- Extra tracing filters (`--tracing-filter`) can be added in tests
- Span fields can now be matched as well as names

### DIFF
--- a/test-harness/src/matcher/span_matcher.rs
+++ b/test-harness/src/matcher/span_matcher.rs
@@ -1,0 +1,64 @@
+use std::fmt::Display;
+
+use crate::tracing_json::Span;
+
+use super::FieldMatcher;
+
+/// A [`Span`] matcher.
+#[derive(Clone)]
+pub struct SpanMatcher {
+    name: String,
+    fields: FieldMatcher,
+}
+
+impl SpanMatcher {
+    /// Construct a query for spans with the given name.
+    pub fn new(name: impl AsRef<str>) -> Self {
+        Self {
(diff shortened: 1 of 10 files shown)


test-harness/src/matcher/span_matcher.rs
```diff
@@ -0,0 +1,64 @@
+use std::fmt::Display;
+
+use crate::tracing_json::Span;
+
+use super::FieldMatcher;
+
+/// A [`Span`] matcher.
+#[derive(Clone)]
+pub struct SpanMatcher {
+    name: String,
```
Gabriella439, line 10: I noticed that for field values you do a regex match but for span names you do an exact match.  I was wondering if it would be possible to also support a regex match for the span name.

Gabriella439 (dismissed):
looks great!  just a couple of non-blocking things

test-harness/src/matcher/span_matcher.rs
9999years: Yeah, I just haven't needed it yet. I also don't want to have to remember to write `^run_hooks$` instead of `run_hooks` every time I use it.

Gabriella439 (approved):

PR #195, opened 2024-02-26
Clear the screen on reload with `--clear`

9999years:
It's possible to hack around this with lifecycle hooks:

```shell
ghciwatch \
  --before-startup-shell 'sh -c "clear >/dev/tty"' \
  --before-reload-shell 'sh -c "clear >/dev/tty"' \
  --before-restart-shell 'sh -c "clear >/dev/tty"'
```

But it's nice to have a baked option for it.


evanrelf (approved):
I liked having the screen cleared the moment I started `ghciwatch`, so it'd be nice to have that here too.

PR #201, opened 2024-02-28
TUI nits

9999years:
A couple minor changes to #185 before we merge it.

### DIFF
--- a/src/buffers.rs
+++ b/src/buffers.rs
@@ -22,3 +22,7 @@ pub const TRACING_BUFFER_CAPACITY: usize = 1024;
 
 /// Size of a buffer for `ghci` output. Used to implement the TUI.
 pub const GHCI_BUFFER_CAPACITY: usize = 1024;
+
+/// Initial capacity for the TUI scrollback buffer, containing data written from `ghci` and
+/// `tracing` log messages.
+pub const TUI_SCROLLBACK_CAPACITY: usize = 16384;
--- a/src/buffers.rs
+++ b/src/buffers.rs
@@ -25,4 +25,4 @@ pub const GHCI_BUFFER_CAPACITY: usize = 1024;
 
 /// Initial capacity for the TUI scrollback buffer, containing data written from `ghci` and
 /// `tracing` log messages.
-pub const TUI_SCROLLBACK_CAPACITY: usize = 16384;
+pub const TUI_SCROLLBACK_CAPACITY: usize = 16 * 1024;
--- a/src/tui/mod.rs
+++ b/src/tui/mod.rs
@@ -46,6 +46,8 @@ pub async fn run_tui(
 
     let mut event_stream = EventStream::new();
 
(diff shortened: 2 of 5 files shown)


src/buffers.rs
```diff
@@ -22,3 +22,7 @@ pub const TRACING_BUFFER_CAPACITY: usize = 1024;
 
 /// Size of a buffer for `ghci` output. Used to implement the TUI.
 pub const GHCI_BUFFER_CAPACITY: usize = 1024;
+
+/// Initial capacity for the TUI scrollback buffer, containing data written from `ghci` and
+/// `tracing` log messages.
+pub const TUI_SCROLLBACK_CAPACITY: usize = 16384;
```
evanrelf, line 28: I'd write this as `16 * 1024` otherwise it seems too magical.

src/tui/terminal.rs
```diff
@@ -32,20 +35,32 @@ impl DerefMut for TerminalGuard {
...
+#[instrument(level = "debug")]
 pub fn enter() -> miette::Result<TerminalGuard> {
     use event::KeyboardEnhancementFlags as KEF;
 
-    INSIDE.store(true, Ordering::SeqCst);
+    if INSIDE.load(Ordering::SeqCst) {
+        return Err(miette!(
+            "Cannot enter raw mode; the terminal is already set up"
+        ));
+    }
```
evanrelf, line 63: Instead of erroring, you could `return Ok(TerminalGuard { .. })` so it's idempotent instead.

src/tui/mod.rs
```diff
@@ -114,6 +121,7 @@ fn render(tui: &Tui, area: Rect, buffer: &mut Buffer) -> miette::Result<()> {
 
 const SCROLL_AMOUNT: usize = 1;
 
+#[instrument(level = "trace", skip(tui))]
 fn handle_event(tui: &mut Tui, event: Event) -> miette::Result<()> {
     match event {
         // TODO(evan): Scrolling is excruciatingly slow
```
evanrelf, line 127: Stale todo now that `gag` is gone

src/tui/terminal.rs
```diff
@@ -83,9 +98,13 @@ pub fn enter() -> miette::Result<TerminalGuard> {
         .into_diagnostic()
         .wrap_err("Failed to create ratatui terminal")?;
 
+    INSIDE.store(true, Ordering::SeqCst);
```
evanrelf, line 101: Hmm... I feel like `INSIDE` should be set ASAP rather than after we're finished doing all the `enter` tasks, so that an incomplete `enter` gets rolled back by `exit`.

evanrelf (approved):

src/tui/mod.rs
9999years: Oh I thought that referred to the scroll speed rather than the event speed.

evanrelf (approved):

PR #206, opened 2024-03-01
Add key bindings in TUI mode

9999years:
Adds a bunch of Vim/`less`-like key bindings I wanted when I was testing the original TUI PR. These will probably be removed as soon as we do anything interesting with the UI, but they're nice for now.

### DIFF
--- a/src/tui/mod.rs
+++ b/src/tui/mod.rs
@@ -133,6 +133,8 @@ fn handle_event(tui: &mut Tui, event: Event, frame: Frame<'_>) -> miette::Result
         .count()
         .saturating_sub(1);
 
+    // TODO: Steal Evan's declarative key matching macros?
+    // https://github.com/evanrelf/indigo/blob/7a5e8e47291585cae03cdf5a7c47ad3bcd8db3e6/crates/indigo-tui/src/key/macros.rs
     match event {
         Event::Mouse(mouse) if mouse.kind == MouseEventKind::ScrollUp => {
             tui.scroll_offset = tui.scroll_offset.saturating_sub(SCROLL_AMOUNT);
--- a/src/tui/mod.rs
+++ b/src/tui/mod.rs
@@ -94,6 +94,8 @@ pub async fn run_tui(
                     .ok_or_else(|| miette!("No more crossterm events"))?
                     .into_diagnostic()
                     .wrap_err("Failed to get next crossterm event")?;
+                // TODO: `get_frame` is an expensive call, delay if possible.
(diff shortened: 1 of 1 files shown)


src/tui/mod.rs
```diff
@@ -90,7 +92,7 @@ pub async fn run_tui(
                     .ok_or_else(|| miette!("No more crossterm events"))?
                     .into_diagnostic()
                     .wrap_err("Failed to get next crossterm event")?;
-                handle_event(&mut tui, event)?;
+                handle_event(&mut tui, event, terminal.get_frame())?;
```
evanrelf, line 95: You should pass the `Rect` representing the size of the `Frame`, rather than the entire `Frame` (which has explicit lifetimes, access to the terminal `Buffer`, etc.).

```suggestion
                handle_event(&mut tui, event, terminal.get_frame().size())?;
```

src/tui/mod.rs
```diff
@@ -121,27 +123,70 @@ fn render(tui: &Tui, area: Rect, buffer: &mut Buffer) -> miette::Result<()> {
...
+                _ => {}
+            },
+
+            #[allow(clippy::single_match)]
+            KeyModifiers::SHIFT => match key.code {
+                KeyCode::Char('G') => {
+                    tui.scroll_offset = last_line;
+                }
+                _ => {}
+            },
```
evanrelf, line 161: You might be interested in the declarative macros I wrote for my text editor to match on `crossterm`'s `Key`s more concisely:

- [Usage example](https://github.com/evanrelf/indigo/blob/7a5e8e47291585cae03cdf5a7c47ad3bcd8db3e6/crates/indigo-tui/src/main.rs#L283-L307)
  ```rust
    match event {
        Event::Key(key) if k!(key, Left) || k!(key, CONTROL, 'b') => {
            command_mode.move_backward(1);
        }
        Event::Key(key) if k!(key, Right) || k!(key, CONTROL, 'f') => {
            command_mode.move_forward(1);
        }
        Event::Key(key) if k!(key, Left) || k!(key, ALT, 'b') => {
            command_mode.move_backward_word();
        }
    // ...
  ```
- [Implementation](https://github.com/evanrelf/indigo/blob/7a5e8e47291585cae03cdf5a7c47ad3bcd8db3e6/crates/indigo-tui/src/key/macros.rs)

evanrelf (approved):

src/tui/mod.rs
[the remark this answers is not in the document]
9999years: I actually have another change to implement some of this! I'll merge this as-is and then fix it up.

PR #208, opened 2024-03-01
Refactor TUI

9999years:
`Tui` is renamed to `TuiState`, `Tui` is now a bundle of a `TerminalGuard` and `TuiState`. This makes the code nicer.

### DIFF
--- a/src/tui/mod.rs
+++ b/src/tui/mod.rs
@@ -170,7 +170,7 @@ impl Tui {
 
                 #[allow(clippy::single_match)]
                 KeyModifiers::SHIFT => match key.code {
-                    KeyCode::Char('G') => {
+                    KeyCode::Char('g' | 'G') => {
                         self.scroll_to(usize::MAX);
                     }
                     _ => {}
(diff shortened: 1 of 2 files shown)


src/tui/mod.rs
```diff
@@ -15,24 +16,196 @@ use ratatui::prelude::Rect;
...
+                    _ => {}
+                },
+
+                #[allow(clippy::single_match)]
+                KeyModifiers::SHIFT => match key.code {
+                    KeyCode::Char('G') => {
+                        self.scroll_to(usize::MAX);
+                    }
+                    _ => {}
+                },
```
evanrelf, line 177: I wonder if there are terminals that report this as `SHIFT` + `g` and not `SHIFT` + `G`? Maybe writing these `SHIFT`-based keys like this would be more robust:

```suggestion
                KeyModifiers::SHIFT => match key.code {
                    KeyCode::Char('g' | 'G') => {
                        self.scroll_to(usize::MAX);
                    }
                    _ => {}
                },
```

evanrelf (approved):

evanrelf (approved):

PR #210, opened 2024-03-01
TUI: Follow output when scrolled to the bottom

### LINKED ISSUE #209: [DUX-1987] TUI: Follow output when scrolled to bottom
<sub>From [SyncLinear.com](https://synclinear.com) | [DUX-1987](https://linear.app/mercury/issue/DUX-1987/tui-follow-output-when-scrolled-to-bottom)</sub>

9999years:
Fixes #209.

### DIFF
--- a/src/tui/mod.rs
+++ b/src/tui/mod.rs
@@ -1,4 +1,4 @@
-use std::cmp::min;
+use std::num::Saturating;
 use std::ops::Deref;
 use std::ops::DerefMut;
 
@@ -36,8 +36,8 @@ const SCROLL_AMOUNT: usize = 3;
 struct TuiState {
     quit: bool,
     scrollback: Vec<u8>,
-    line_count: usize,
-    scroll_offset: usize,
+    line_count: Saturating<usize>,
+    scroll_offset: Saturating<usize>,
 }
 
 impl Default for TuiState {
@@ -45,8 +45,8 @@ impl Default for TuiState {
         Self {
             quit: false,
             scrollback: Vec::with_capacity(TUI_SCROLLBACK_CAPACITY),
-            line_count: 1,
-            scroll_offset: 0,
+            line_count: Saturating(1),
+            scroll_offset: Saturating(0),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/tui/mod.rs
```diff
@@ -122,10 +122,18 @@ impl Tui {
         self.scroll_offset = min(self.scroll_max(), scroll_offset);
     }
 
+    fn maybe_follow(&mut self) {
+        let height = self.size.height as usize;
+        if self.scroll_offset >= self.line_count - height - 1 {
```
evanrelf, line 127: Nit pick: I would add parens to make order of operations more obvious.

```suggestion
        if self.scroll_offset >= self.line_count - (height + 1) {
```

evanrelf, line 127: What if `height` is greater than `line_count` (e.g. when first starting)? Might need some saturating arithmetic or some other check here.

src/tui/mod.rs
```diff
@@ -122,10 +122,18 @@ impl Tui {
...
+            self.scroll_offset += 1;
```
evanrelf, line 128: I think you're using `saturating_add` in other places, but not here?

evanrelf (changes requested):

src/tui/mod.rs
```diff
@@ -36,17 +36,17 @@ const SCROLL_AMOUNT: usize = 3;
...
 }
 
 impl Default for TuiState {
     fn default() -> Self {
         Self {
             quit: false,
             scrollback: Vec::with_capacity(TUI_SCROLLBACK_CAPACITY),
-            line_count: 1,
-            scroll_offset: 0,
+            line_count: Saturating(1),
```
evanrelf, line 48: lol you could go even further and make this a `NonZeroUsize`

9999years: `NonZeroUsize` isn't saturating, weirdly: https://doc.rust-lang.org/stable/core/num/struct.NonZeroUsize.html#method.saturating_add

PR #216, opened 2024-04-11
Support multiple spans in `BaseMatcher::in_leaf_span`

9999years:
This makes it easier to identify the correct span (e.g. `in_spans(["run_ghci", "initialize"])` would match an event with spans `["run_ghci", "stdout", "initialize"]`, but `in_leaf_spans` would not).

### DIFF
--- a/test-harness/src/matcher/base_matcher.rs
+++ b/test-harness/src/matcher/base_matcher.rs
@@ -14,7 +14,7 @@ use super::SpanMatcher;
 pub struct BaseMatcher {
     message: Regex,
     target: Option<String>,
-    leaf_span: Option<SpanMatcher>,
+    leaf_spans: Vec<SpanMatcher>,
     spans: Vec<SpanMatcher>,
     fields: FieldMatcher,
 }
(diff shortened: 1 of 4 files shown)


test-harness/src/matcher/base_matcher.rs
```diff
@@ -214,13 +225,17 @@ impl Matcher for BaseMatcher {
...
         }
 
-        if let Some(leaf_span_matcher) = &self.leaf_span {
-            match event.spans().last() {
-                Some(leaf_span) if leaf_span_matcher.matches(leaf_span) => {}
-                _ => {
-                    // Expected a leaf span, but the event is missing a leaf span or doesn't have
-                    // the correct leaf span.
-                    return Ok(false);
+        if !self.leaf_spans.is_empty() {
```
Gabriella439, line 228: Is this check necessary?  I think the code behaves the same without it (since the `for` loop will do nothing if `self.leaf_spans` is empty

9999years: Good point. Removed this check (and the one for the loop above it).

Gabriella439 (approved):

PR #223, opened 2024-04-30
Restore Rust 1.72-ability

9999years:


evanrelf (approved):
Is this to avoid depending on the reverted Nixpkgs upgrade in `mwb`?

9999years:
Yeah 😫

PR #227, opened 2024-05-01
Add `--read-logs-from` for asynchronous log output

9999years:
Because ghciwatch reads ghci's output and expects it to be structured in a standard way, it can't cope with programs that output to stdout (or stderr) asynchronously (see https://github.com/ndmitchell/ghcid/issues/137).

To make this easier for frameworks that output logs (like Yesod), we should have an option to read logs asynchronously from a file. If provided, writes to the given file will show up in ghciwatch's output (but not ghci's output).


src/read_logs_from.rs
```diff
@@ -0,0 +1,148 @@
...
+            ..Default::default()
+        };
+
+        let mut stdout = tokio::io::stdout();
+
+        while let Some(duration) = backoff.next_backoff() {
+            while let Some(line) = lines.next_line().await.into_diagnostic()? {
+                // TODO: Lock stdout here and for ghci output.
+                let _ = stdout.write_all(line.as_bytes()).await;
+                let _ = stdout.write_all(b"\n").await;
```
9999years, line 119: This is a pretty big "TODO". `std::io::Stdout` is globally synchronized, but [`tokio::io::Stdout`](https://docs.rs/tokio/latest/tokio/io/struct.Stdout.html) is not, so this could lead to interleaved output.

AFAICT there's no easy solution for this. The tokio devs suggested creating a task to manage stdout and sending it messages (strings to write) over a channel, which is reasonable enough, but then I'll have to write an adapter to implement [`std::io::Write`](https://doc.rust-lang.org/stable/std/io/trait.Write.html) by sending messages to an async channel.

src/main.rs
```diff
@@ -45,6 +46,18 @@ async fn main() -> miette::Result<()> {
...
     }
 
+    for path in opts.read_logs_from {
+        manager
+            .spawn("read-logs", |handle| {
+                ReadLogsFrom {
+                    shutdown: handle,
+                    path,
+                }
+                .run()
```
Gabriella439, line 56: This is not a blocker, but it wasn't clear to me why this is split up as a struct and a method on that struct (i.e. `ReadLogsFrom { shutdown: handle, path }.run()`) and not combined into a single function (i.e. `readLogsFrom(handle, path)`)

Gabriella439 (approved):

src/main.rs
9999years: A couple reasons why I chose to implement it like this:

- Struct+method is a lot easier to split up to deal with complex logic.
- Methods don't have named parameters, so when there's a lot of state to pass in, using a struct is a lot clearer (and lower-overhead than defining builders).
- Most of the async tasks in `ghciwatch` are structured like this.

PR #237, opened 2024-05-15
Fix "module defined in multiple files" AGAIN!!!!

9999years:
In https://github.com/MercuryTechnologies/ghciwatch/pull/214, we had a situation where modules were loaded:

    ghci> :show targets
    Foo
    Bar
    Baz

And then an eval comment in (e.g.) `Foo` causes the module to be added and explicitly interpreted by path:

    ghci> :add *src/Foo.hs

Then, we have `Foo` loaded by name (`Foo`) and by path (`src/Foo.hs`), which triggers the dreaded bug.

At the time I proposed this fix, correctly:

> I think we can fix this by keeping track of how each module is added to the session — as a path or as a module name — and then only using that form going forward.

I threaded some extra information to the `:show targets` parser to track if modules were listed as names or paths, **but then at the end of `Ghci::interpret_module` I would always insert the module into the target set as a `TargetKind::Path`,** meaning that the *next* time the comment was evaluated, the module would be loaded as a path, causing the
error.

https://github.com/MercuryTechnologies/ghciwatch/blob/dbba61bbdec9a86f97051b12647e51b7be4fd484/src/ghci/mod.rs#L698-L699

This fixes the bug and adds an `assert!` to fail faster and more obviously if it happens again.

## Prior art

- #37
- #77 
- #125 
- #214

### DIFF
--- a/src/ghci/parse/module_set.rs
+++ b/src/ghci/parse/module_set.rs
@@ -52,7 +52,13 @@ impl ModuleSet {
     ///
     /// Returns whether the value was newly inserted.
     pub fn insert_source_path(&mut self, path: NormalPath, kind: TargetKind) -> bool {
-        self.modules.insert(path, kind).is_some()
+        match self.modules.insert(path, kind) {
+            Some(old_kind) => {
+                assert!(kind == old_kind, "`ghciwatch` failed to track how modules were imported in `ghci`; please report this as a bug");
+                true
+            }
+            None => false,
+        }
     }
 
     /// Get the name used to refer to the given module path when importing it.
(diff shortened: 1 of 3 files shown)


src/ghci/parse/module_set.rs
```diff
@@ -89,3 +104,16 @@ impl ModuleSet {
...
     }
 }
+
+/// Information about a module to be imported into a `ghci` session.
+pub struct ImportInfo {
+    /// The name to refer to the module by.
+    ///
+    /// This may either be a dotted module name like `My.Cool.Module` or a path like
+    /// `src/My/Cool/Module.hs`.
+    pub name: String,
```
Gabriella439, line 114: Minor thing, but I think making the `String` a field for each of the `Path` and `Module` constructors for `TargetKind` would make it less likely to accidentally confuse the two types of `String`s (i.e. using a module where a path was expected or vice versa).  However, I realize that's a more extensive refactor

Gabriella439 (approved):

PR #247, opened 2024-05-20
Add vendored/forked clap-markdown

9999years:
This will generate CLI documentation for ghciwatch's user manual.

See the rendered docs: https://github.com/MercuryTechnologies/ghciwatch/blob/rebeccat/dux-2305-cli-documentation/clap-markdown/examples/complex_app.md

PR #248, opened 2024-05-20
Integrate `clap-markdown` in `ghciwatch`

9999years:
Blocked on #247


src/cli.rs
```diff
@@ -14,10 +12,47 @@ use crate::clonable_command::ClonableCommand;
 use crate::ignore::GlobMatcher;
 use crate::normal_path::NormalPath;
 
-/// A `ghci`-based file watcher and Haskell recompiler.
+/// Ghciwatch loads a GHCi session for a Haskell project and reloads it
+/// when source files change.
+///
+/// ## Examples
```
9999years, line 18: Adding some examples. This is a sort of weird compromise; these look decent in `--help` output unprocessed and also render as Markdown.

`clap` doesn't have any hooks to customize how the docstrings are preprocessed, so we have to turn off preprocessing (or it'll line-wrap the examples). Unfortunately, that loses us all formatting for this text...

src/cli.rs
```diff
@@ -14,10 +12,47 @@ use crate::clonable_command::ClonableCommand;
...
+///
+/// Also reload the session when `.persistentmodels` change:
+///
+///     ghciwatch --watch config/modelsFiles \
+///               --reload-glob '**/*.persistentmodels'
+///
+/// Don't reload for `README.md` files:
+///
+///     ghciwatch --reload-glob '!src/**/README.md'
+#[allow(rustdoc::invalid_rust_codeblocks)]
```
9999years, line 47: And `rustdoc` doesn't like that the indented shell examples aren't valid Rust code. (You're supposed to use triple backticks annotated as `plain` or something for that, but of course we don't want it to show up in the output.)

src/cli.rs
```diff
@@ -14,10 +12,47 @@ use crate::clonable_command::ClonableCommand;
...
 #[derive(Debug, Clone, Parser)]
-#[command(version, author, about)]
-#[command(max_term_width = 100)]
+#[command(
+    version,
+    author,
+    verbatim_doc_comment,
+    max_term_width = 100,
+    override_usage = "ghciwatch [--command SHELL_COMMAND] [--watch PATH] [OPTIONS ...]"
```
9999years, line 54: `--command` and `--watch` are the really important options so I pulled them out; the default usage string was `ghciwatch [OPTIONS]` which isn't very useful.

src/cli.rs
```diff
@@ -88,33 +132,40 @@ pub struct WatchOpts {
...
-    /// Restart the `ghci` session when paths matching this glob change. Can be given multiple
-    /// times.
+    /// Restart the `ghci` session when paths matching this glob change.
     ///
     /// By default, only changes to `.cabal` or `.ghci` files or Haskell source files being
     /// moved/removed will trigger restarts.
     ///
-    /// Due to a `ghci` bug, the `ghci` session must be restarted when Haskell modules are removed
-    /// or renamed: https://gitlab.haskell.org/ghc/ghc/-/issues/11596
+    /// Due to [a `ghci` bug][1], the `ghci` session must be restarted when Haskell modules are removed
```
9999years, line 161: Kinda clumsy to have markdown link syntax in the `--help` output like this but it looks OK enough and it renders correctly as Markdown.

It'd be nice to have a system that let us do some more advanced templating to add more detail in the manual...

Gabriella439 (approved):

PR #256, opened 2024-05-20
Run after-restart hooks before eval commands and test suites

9999years:
Fixes an inconsistency between after-startup and after-restart hooks

### DIFF
--- a/src/ghci/mod.rs
+++ b/src/ghci/mod.rs
@@ -343,7 +343,11 @@ impl Ghci {
     ///
     /// Diagnostics will be added to the given `log`, and the error log will be written.
     #[instrument(level = "debug", skip_all)]
-    pub async fn initialize(&mut self, log: &mut CompilationLog) -> miette::Result<()> {
+    pub async fn initialize<const N: usize>(
+        &mut self,
+        log: &mut CompilationLog,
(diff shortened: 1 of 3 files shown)


src/ghci/mod.rs
```diff
@@ -767,16 +768,20 @@ impl Ghci {
...
     ) -> miette::Result<()> {
         // Allow hooks to consume the error log by updating it before running the hooks.
         self.write_error_log(log).await?;
 
-        self.run_hooks(event, log).await?;
+        for event in events {
+            self.run_hooks(event, log).await?;
+        }
+
+        let event = events[N - 1];
```
Gabriella439, line 784: Is there a way to enforce that `N` is greater than 0?

Alternatively, another way to guarantee the same thing is to split the `events` argument into two arguments: `events: [LifecycleEvent]` (for all but the last event) and `event: LifecycleEvent` (for the very last event).  This is essentially a poor person's encoding of a non-empty list.

I don't feel strongly about this, though.

Gabriella439 (approved):

src/ghci/mod.rs
9999years: Not possible yet apparently :(

PR #288, opened 2024-06-05
Adding a small test for normal_path.rs

atruslow:
- Adding a small test for normal_path.rs

### DIFF
--- a/src/normal_path.rs
+++ b/src/normal_path.rs
@@ -191,7 +191,7 @@ mod tests {
     use super::*;
 
     #[test]
-    fn test_into_relative() {
+    fn test_into_absolute() {
         let dir = env::temp_dir();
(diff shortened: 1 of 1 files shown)


src/normal_path.rs
```diff
@@ -183,3 +183,22 @@ impl ValueParserFactory for NormalPath {
...
+
+#[cfg(test)]
+mod tests {
+    use std::env;
+
+    use super::*;
+
+    #[test]
+    fn test_into_absolute() {
+        let dir = env::temp_dir();
```
9999years, line 195: I suspect `env::temp_dir()` is already absolute, so I'm not sure how useful this is as a test. Looks good though.

9999years (approved):

PR #290, opened 2024-06-07
Don't restart for removed or renamed modules

9999years:
Ghcid crashes when modules are removed/renamed because all it can do to the underlying GHCi session is `:reload`, and [a GHCi bug][1] means that the GHCi session will crash when this happens.

Because ghciwatch tracks which files/modules are loaded in the session, it can accurately `:unadd` files from the session when they're removed on disk, without requiring the session be restared.

This should speed up development.

[1]: https://gitlab.haskell.org/ghc/ghc/-/issues/11596

I'm going to cut 1.0 with this while we're at it.

- [x] Labeled the PR with `patch`, `minor`, or `major` to request a version bump when it's merged.


src/ghci/mod.rs
```diff
@@ -707,6 +718,28 @@ impl Ghci {
...
     }
 
+    /// `:unadd` a module or modules from the `ghci` session by path.
+    #[instrument(skip(self), level = "debug")]
+    async fn remove_modules(
+        &mut self,
+        paths: &[NormalPath],
+        log: &mut CompilationLog,
+    ) -> miette::Result<()> {
+        let modules = self.targets.format_modules(&self.search_paths, paths)?;
```
9999years, line 728: Each `:unadd` implicitly reloads as well, so we have to `:unadd` all the modules in a single command so that GHCi doesn't try to load a bunch of removed modules after each one.

tests/remove.rs
```diff
@@ -0,0 +1,40 @@
...
+async fn can_remove_multiple_modules_at_once() {
+    let mut session = GhciWatch::new("tests/data/simple")
+        .await
+        .expect("ghciwatch starts");
+    session
+        .wait_until_ready()
+        .await
+        .expect("ghciwatch loads ghci");
+
+    session.fs_mut().disable_load_bearing_sleep();
```
9999years, line 15: This API is clumsy, the load bearing sleep value should be tied to the particular `Fs` instance, not shared.

src/ghci/parse/module_set.rs
```diff
@@ -99,13 +111,32 @@ impl ModuleSet {
...
+    /// See [`ModuleSet::module_import_name`].
+    pub fn format_modules(
+        &self,
+        show_paths: &ShowPaths,
+        modules: &[NormalPath],
+    ) -> miette::Result<String> {
+        modules
+            .iter()
+            .map(|path| {
+                self.module_import_name(show_paths, path)
```
9999years, line 125: These `ModuleSet` APIs are showing some strain, but it's not clear what to replace them with. I tried a few designs and didn't like the results.

src/ghci/mod.rs
Gabriella439: What you said should probably be a comment in the code

tests/globs.rs
```diff
@@ -182,12 +182,41 @@ async fn can_restart_on_module_change_even_if_ignored() {
...
         .unwrap();
 
     session
-        .wait_until_restart()
+        .wait_for_log(BaseMatcher::reload_completes().but_not(BaseMatcher::ghci_remove()))
+        .await
+        .expect("Ghciwatch ignores removed files");
+}
+
+/// Ghciwatch can whiteslist reloads for removed files.
```
Gabriella439, line 195:
```suggestion
/// Ghciwatch can whitelist reloads for removed files.
```

Gabriella439 (dismissed):

atruslow (approved):
LGTM!

PR #292, opened 2024-06-12
`ModuleSet`: store a set rather than a map

9999years:
Previously, a `ModuleSet` wrapped a `HashMap<NormalPath, TargetKind>`. This had a number of undesirable consequences:
* The data about a module's name (as its path) and how it was loaded (as its `TargetKind`) were split from each other and difficult to reference.
* The module's import name wasn't explicitly stored anywhere, so we needed to convert between paths and dotted names when those were needed, which required hitting the disk.
* There wasn't a type for the module's import name, so when we (e.g.) `:unadd`ed modules we needed to format them as strings.

Now, a `ModuleSet` wraps a `HashSet<LoadedModule>`.

* A `LoadedModule` wraps a path but optionally contains the module's dotted name, if the module is loaded by name (and needs to be referred to by name to avoid the "module defined in multiple files" error).
* The `LoadedModule` `Display` instance formats the module's import name correctly (with a dotted name if needed) and avoids hitting the disk or any string processing.


src/ghci/loaded_module.rs
```diff
@@ -0,0 +1,96 @@
...
+        Self { path, name: None }
+    }
+
+    /// Create a new module, loaded by name.
+    pub fn with_name(path: NormalPath, name: String) -> Self {
+        Self {
+            path,
+            name: Some(name),
+        }
+    }
```
Gabriella439, line 42: I think I might have asked this before, but what's the motivation for using these methods instead of using the `LoadedModule` constructor directly?

src/ghci/loaded_module.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    /// Get the module's source path.
+    pub fn path(&self) -> &NormalPath {
+        &self.path
```
Gabriella439, line 46: Similar question here: what's the advantage of doing `loadedModule.path()` instead of `loadedModule.path`?

src/ghci/module_set.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::normal_path::NormalPath;
+
+use super::loaded_module::LoadedModule;
+
+/// A collection of source paths, retaining information about loaded modules in a `ghci`
+/// session.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ModuleSet {
+    modules: HashSet<LoadedModule>,
+}
```
Gabriella439, line 16: Non-blocking, but I feel like this `struct` wrapper around the `HashSet` doesn't buy you much because most of the methods on `ModuleSet` are trivially wrapping methods on the underlying `HashSet`.  You could probably pass around the `HashSet` directly instead of defining a separate `ModuleSet` `struct`.

Gabriella439 (approved):

src/ghci/loaded_module.rs
9999years: It's the usual encapsulation reason. If you want to be able to write `LoadedModule { path, name: None }` then all of the struct's fields have to be public (and theoretically callers could use this to break invariants).

In reality, `LoadedModule` isn't exported publicly by the library and it would be basically fine to have the fields be public, but this is considered "good practice" so I lean towards it.

(Also constructor names can be a little more informative than setting struct fields directly.)

src/ghci/module_set.rs
9999years: Yeah, that's true. I like the type safety this gives, though, and I like being able to give semantically descriptive names to the methods. I think I'll keep this for now.

9999years:
I tried making the `LoadedModule` fields public and replacing `ModuleSet` with `HashSet<LoadedModule>`. The diff was +67 -148 (net -81 lines). I like the type safety and the implementation of `ModuleSet` has changed before, so I'm inclined to keep it at least for now.

PR #313, opened 2024-08-02
Update README.md

theGhostJW:
The manual looks really good and deserves a spot in the readme

- [ x ] `minor`
- [ x ] Updated the user manual in `docs/`.


9999years:
Thanks! There actually is a badge linking to the manual but I guess it's nestled in with the versions of ghciwatch in various package managers so it's easy to miss:

image

9999years (approved):

Issue #317, opened 2024-08-15
Reloading after filechange no longer works when setting `hSetBuffering stdout LineBuffering` in ghci, affecting Tasty tests.


erikryb:
### What happened?

Ghciwatch stops reloading after file changes if the command `hSetBuffering stdout LineBuffering` is ran in ghci. This affects running tests with Tasty.

### What did you expect to happen?

Ghciwatch should reload after file changes.

### Steps to reproduce the issue

Create a project with one file:

```
module Main where

-- $> hSetBuffering stdout LineBuffering
```

Run `ghciwatch --enable-eval`. Edit the file and save. ghciwatch does not reload. If ghciwatch is aborted with CTRL-C, we get the following error:

```
^CError:   × Tasks failed:
  │ • run_ghci: ghci event channel closed
```

### The version of ghciwatch with the bug

ghciwatch 1.0.1

erikryb:
I have found a workaround, using a wrapper function: 

```
bubblewrap :: IO () -> IO ()
bubblewrap io = do
  try io :: IO (Either SomeException ())
  hSetBuffering stdout NoBuffering
```

This wrapper function takes an IO action, and returns a new IO action that can be run by ghciwatch. It handles any exception (which is needed if the IO action performs an exit, which is the case for Tasty) and also sets the buffering back to NoBuffering after performing the IO action. We can now run Tasty tests in ghciwatch by adding the following comment in the source file:

`-- $> bubblewrap (defaultMain =<< tests)`

and running `ghciwatch --enable-eval`.

9999years:
Yeah, don't do that. The wrapper is a good solution. If stdout is line-buffered, then ghciwatch can't detect when GHCi writes a prompt (because the prompt is written without a trailing newline).

PR #321, opened 2024-08-30
add tasty integration notes

mwotton:
- [ ] Labeled the PR with `patch`, `minor`, or `major` to request a version bump when it's merged. (N/A, docs-only)
- [ ] Added integration / regression tests in `tests/`. (N/A, docs-only)

### DIFF
--- a/docs/tasty-integration.md
+++ b/docs/tasty-integration.md
@@ -1,0 +1,28 @@
+# using tasty with ghciwatch
+
+## bubblewrap
+
+Because `ghciwatch` is waiting for lines to come from `ghci`, if you don't change the buffering, you
+can end up waiting forever if you don't change the output buffering from `Tasty`. Something like this works:
+
+```haskell
+module TestMain where
+
+import Control.Exception (SomeException, try)
+import System.IO (BufferMode (NoBuffering), hSetBuffering, stdout)
+import Test.Tasty (TestTree, defaultMain, testGroup)
+
+bubblewrap :: IO () -> IO ()
+bubblewrap io = do
+  try io :: IO (Either SomeException ())
+  hSetBuffering stdout NoBuffering
+
+main :: IO ()
+main = bubblewrap $ defaultMain $ mytestgroup
+```
+
+## tasty-discover issues
+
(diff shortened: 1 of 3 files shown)


mwotton:
thanks to @erikryb for the patch https://github.com/MercuryTechnologies/ghciwatch/issues/317

docs/tasty-integration.md
```diff
@@ -0,0 +1,28 @@
...
+
+main :: IO ()
+main = bubblewrap $ defaultMain $ mytestgroup
+```
+
+## tasty-discover issues
+
+If you add a new test file, the top level [tasty-discover](https://hackage.haskell.org/package/tasty-discover)
+module will not have it set up as a dependency, so will not be reloaded until you restart the `ghciwatch` process. [tasty-autocollect](https://hackage.haskell.org/package/tasty-autocollect) relies on a compiler plugin and seems to avoid 
+this problem.
```
parsonsmatt, line 28: This *shouldn't* be true - if `ghciwatch` sees an edit to the file that uses `tasty-discover` (or any other file that is a dependency of that module), then the GHC preprocessor plugin will run and include the newly created test file. You shouldn't need to restart ghciwatch.

docs/tasty-integration.md
```diff
@@ -0,0 +1,28 @@
...
+module TestMain where
+
+import Control.Exception (SomeException, try)
+import System.IO (BufferMode (NoBuffering), hSetBuffering, stdout)
+import Test.Tasty (TestTree, defaultMain, testGroup)
+
+bubblewrap :: IO () -> IO ()
+bubblewrap io = do
+  try io :: IO (Either SomeException ())
+  hSetBuffering stdout NoBuffering
```
parsonsmatt, line 18: Is tasty changing the buffering mode? I'm curious if the benefit here is to set the `NoBuffering` mode, or if the benefit is just that we're flushing the buffer

9999years (approved):

9999years:
OK, I rewrote this a little. I'm going to merge it now, let me know if you don't like any of my changes and we can sort it out. Also please try writing the top-level `Spec.hs` and see if that helps pick up new test files.

PR #337, opened 2025-01-20
[DUX-2822] Add `--verbose`, an alias for `--log-filter ghciwatch=debug,info`.

a-02:

I don't work here anymore, but this is still bugging me. Is this close to what you were wanting?

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -82,6 +82,10 @@ pub struct Opts {
     #[arg(long, alias = "allow-eval")]
     pub enable_eval: bool,
 
+    /// More verbose logging output.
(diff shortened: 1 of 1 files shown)


9999years:
Unfortunately the `clap` `alias` functionality is pretty limited -- it can only make add aliases for an existing argument (I doubt this works as written, but it would make `----log-filter ghciwatch=debug,info` work the same as `--verbose` if it did).

The actual code won't be that much more complex, we just need to create some logic for figuring out the log filters and then use that. It'll look something like this:

https://github.com/home-mangler/home-mangler/blob/a9faf62e8ebe5ebbe1950daf2fef8ef6dd64eef4/src/cli.rs#L68-L86

PR #338, opened 2025-02-04
GhciEvent: rename to WatcherEvent

lf-:
This is clearer on its purpose: ghci itself does not generate these events, rather they are events to send to ghci.


9999years (approved):
reasonable

PR #339, opened 2025-02-05
Add `$GHCIWATCH_LOG` for `--log-filter`

9999years:


lf- (approved):
clap should shove this in the docs anyway so w/e

PR #344, opened 2025-02-05
Update `flake.lock`, Rust toolchain, GHC

9999years:
- Update `flake.lock` to update `nixpkgs` and other inputs
- Update Rust toolchain from 1.72.1 to 1.81.0

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "advisory-db": {
       "flake": false,
       "locked": {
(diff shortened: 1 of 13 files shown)


lf-:
should be rustc 1.81.1, other than that, should be ok

9999years:
Might be blocked on https://github.com/NixOS/nixpkgs/issues/367686?

PR #345, opened 2025-02-05
[DUX-3004]: handle evil clients that send ANSI escapes after NL

lf-:
Wew. That was a bug.

The reproducer is here: https://github.com/rampion/ghciwatch-bug-demo

After tracing `consume_str`, I found out that the cause of our problems is that hspec includes the NL in their coloured output before resetting colour.

https://github.com/hspec/hspec/blob/fb9916f07e580f2572220d381e1886e765df95b7/hspec-core/src/Test/Hspec/Core/Formatters/Internal.hs#L297-L306

That means that it is a pain in the ass to parse, since it shoves random ANSI escapes onto the next line. The easiest, and probably most correct, solution to this is to just eat ANSI escapes while looking for markers.

- [x] Labeled the PR with `patch`, `minor`, or `major` to request a version bump when it's merged.

### DIFF
--- a/src/incremental_reader.rs
+++ b/src/incremental_reader.rs
@@ -187,7 +187,13 @@ where
                     ret = match ret {
                         Some(lines) => Some(lines),
                         None => {
-                            match opts.find(opts.end_marker, &self.line) {
+                            // HACK(jadel): eat ANSI escapes right before doing our end
+                            // marker matching. This is because hspec is a naughty
+                            // library and outputs "\n\x1b[0mMARKER" which is really
+                            // rude of it imo.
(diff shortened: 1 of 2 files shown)


lf-:
I probably should shove an integration test in there.

9999years (approved):
I guess this is fine? We could also relax the constraint that matches happen at the start of the line, but that's probably a more invasive change. I do wonder if there's other libraries that don't write a trailing newline though.

Actually, I just checked, and `putStr "puppy"` has the same problem. We probably do need to just fix this so it doesn't require the prompt is on the start of the line.

Happy to merge this as a first-line bugfix, but we should do a more complete fix soon (#346).

PR #358, opened 2025-03-08
Support GHC 9.10, 9.12

9999years:
Needs #344

### DIFF
--- a/src/ghci/parse/ghc_message/compilation_summary.rs
+++ b/src/ghci/parse/ghc_message/compilation_summary.rs
@@ -1,3 +1,5 @@
+use std::fmt::Display;
+
 use winnow::ascii::digit1;
 use winnow::combinator::alt;
 use winnow::combinator::opt;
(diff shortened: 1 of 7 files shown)


src/ghci/parse/ghc_message/compilation_summary.rs
```diff
@@ -51,13 +101,10 @@ pub fn compilation_summary(input: &mut &str) -> PResult<CompilationSummary> {
     .parse_next(input)?;
     let _ = " module".parse_next(input)?;
     let _ = opt("s").parse_next(input)?;
-    let _ = " loaded.".parse_next(input)?;
-    let _ = line_ending_or_eof.parse_next(input)?;
+    let _ = ' '.parse_next(input)?;
+    let _ = alt(("loaded", "reloaded", "added", "unadded", "checked")).parse_next(input)?;
```
9999years, line 105: GHC 9.10 says `Ok, 10 modules reloaded` instead of `Ok, 10 modules loaded`. There's some other special cases I've accounted for here, but that was the big one causing problems.

PR #366, opened 2025-03-13
Parse exceptions during reloads

9999years:
The reload summary ("Ok, 10 modules loaded") is still responsible for whether or not "All good!" is printed, so this still needs some work. We could just check if any exceptions (or errors?) are printed and mark it as a compilation failure if so?

### DIFF
--- a/src/ghci/parse/ghc_message/exception.rs
+++ b/src/ghci/parse/ghc_message/exception.rs
@@ -1,0 +1,54 @@
+use winnow::PResult;
+use winnow::Parser;
+
+use crate::ghci::parse::lines::until_newline;
+
+use super::GhcMessage;
+
+/// Parse an "Exception" message like this:
+///
+/// ```plain
+/// *** Exception: /Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)
+/// ```
+pub fn exception(input: &mut &str) -> PResult<GhcMessage> {
+    let _ = "*** Exception: ".parse_next(input)?;
+
+    let message = until_newline.parse_next(input)?;
+
+    Ok(GhcMessage::Exception(message.to_owned()))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use indoc::indoc;
+    use pretty_assertions::assert_eq;
+
+    #[test]
+    fn test_parse_exception() {
+        assert_eq!(
+            exception.parse("*** Exception: Uh oh!\n").unwrap(),
+            GhcMessage::Exception("Uh oh!".into())
+        );
+
+        assert_eq!(
+            exception.parse("*** Exception: /Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)\n").unwrap(),
+            GhcMessage::Exception("/Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)".into())
+        );
+
+        // Doesn't parse subsequent lines (even if they're relevant, unfortunately).
+        assert_eq!(
+            exception
+                .parse(indoc!(
+                    "
+                    *** Exception: puppy doggy
+                    CallStack (from HasCallStack):
+                      error, called at <interactive>:3:1 in interactive:Ghci1
+                    "
+                ))
+                .unwrap(),
+            GhcMessage::Exception("puppy doggy".into())
+        );
+    }
+}
--- a/src/ghci/parse/ghc_message/mod.rs
+++ b/src/ghci/parse/ghc_message/mod.rs
@@ -44,6 +44,9 @@ use module_import_cycle_diagnostic::module_import_cycle_diagnostic;
 mod no_location_info_diagnostic;
 use no_location_info_diagnostic::no_location_info_diagnostic;
 
+mod exception;
+use exception::exception;
+
 use super::rest_of_line;
 use super::CompilingModule;
 
@@ -64,6 +67,15 @@ pub enum GhcMessage {
     /// Foo.hs:81:1: Warning: Defined but not used: `bar'
     /// ```
     Diagnostic(GhcDiagnostic),
+    /// An exception while running a command or similar.
+    ///
+    /// These may be multiple lines, but there's no good way to determine where the message ends,
+    /// so we just parse the first line.
+    ///
+    /// ```text
+    /// *** Exception: /Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)
+    /// ```
+    Exception(String),
     /// A configuration file being loaded.
     ///
     /// ```text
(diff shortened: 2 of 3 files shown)


src/ghci/parse/ghc_message/exception.rs
```diff
@@ -0,0 +1,54 @@
...
+
+/// Parse an "Exception" message like this:
+///
+/// ```plain
+/// *** Exception: /Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)
+/// ```
+pub fn exception(input: &mut &str) -> PResult<GhcMessage> {
+    let _ = "*** Exception: ".parse_next(input)?;
+
+    let message = until_newline.parse_next(input)?;
```
evanrelf, line 16: I think the most correct parser of a GHCi exception would parse until a prompt, not a newline. But until a newline would handle this particular exception, if that's all you're going for right now.

src/ghci/parse/ghc_message/exception.rs
```diff
@@ -0,0 +1,54 @@
...
+                .parse(indoc!(
+                    "
+                    *** Exception: puppy doggy
+                    CallStack (from HasCallStack):
+                      error, called at <interactive>:3:1 in interactive:Ghci1
+                    "
+                ))
+                .unwrap(),
+            GhcMessage::Exception("puppy doggy".into())
+        );
```
evanrelf, line 52: Ah okay you already know that. Glad to see it still works in the presence of multi-line exception messages?

src/ghci/parse/ghc_message/mod.rs
```diff
@@ -64,6 +67,15 @@ pub enum GhcMessage {
...
     Diagnostic(GhcDiagnostic),
+    /// An exception while running a command or similar.
+    ///
+    /// These may be multiple lines, but there's no good way to determine where the message ends,
+    /// so we just parse the first line.
+    ///
+    /// ```text
+    /// *** Exception: /Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)
+    /// ```
+    Exception(String),
```
evanrelf, line 78: If you don't plan to mutate this string, you could use less memory:

```suggestion
    Exception(Box<str>),
```

If you think `GhcMessage`s will be cloned a lot, you could account for that too:

```suggestion
    Exception(Arc<str>),
```

evanrelf (approved):

src/ghci/parse/ghc_message/exception.rs
9999years: Yeah, the issue is we get these in a stream of arbitrary GHC output, which may include other diagnostics we want to parse, compiling-such-and-such-a-module messages, etc.

src/ghci/parse/ghc_message/mod.rs
9999years: Yeah I'm pretty sure zero-copy parsing is kind of an ergonomic nightmare, even if it's technically possible

evanrelf: > the ergonomics of dealing with both of those types are worse

Not true. All of the `&str` API is still available thanks to the `Deref` impl for `Box<T>` and `Arc<T>`.

> zero-copy parsing is kind of an ergonomic nightmare

This is not zero-copy, and this would not require adding lifetimes.

---

I don't feel strongly about which type you choose, but I do wanna make sure you understand the tradeoffs correctly! I am not suggesting any ergonomic sacrifice here!

This video explains things better: https://www.youtube.com/watch?v=A4cKi7PTJSs

evanrelf: Here's all you'd need to change:


<details>

<summary>Diff</summary>

```diff
diff --git a/src/ghci/parse/ghc_message/exception.rs b/src/ghci/parse/ghc_message/exception.rs
index b09906d..55f4222 100644
--- a/src/ghci/parse/ghc_message/exception.rs
+++ b/src/ghci/parse/ghc_message/exception.rs
@@ -1,3 +1,4 @@
+use std::sync::Arc;
 use winnow::PResult;
 use winnow::Parser;
 
@@ -15,7 +16,7 @@ pub fn exception(input: &mut &str) -> PResult<GhcMessage> {
 
     let message = until_newline.parse_next(input)?;
 
-    Ok(GhcMessage::Exception(message.to_owned()))
+    Ok(GhcMessage::Exception(Arc::from(message)))
 }
 
 #[cfg(test)]
@@ -29,12 +30,12 @@ mod tests {
     fn test_parse_exception() {
         assert_eq!(
             exception.parse("*** Exception: Uh oh!\n").unwrap(),
-            GhcMessage::Exception("Uh oh!".into())
+            GhcMessage::Exception(Arc::new("Uh oh!"))
         );
 
         assert_eq!(
             exception.parse("*** Exception: /Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)\n").unwrap(),
-            GhcMessage::Exception("/Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)".into())
+            GhcMessage::Exception(Arc::new("/Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)"))
         );
 
         // Doesn't parse subsequent lines (even if they're relevant, unfortunately).
@@ -48,7 +49,7 @@ mod tests {
                     "
                 ))
                 .unwrap(),
-            GhcMessage::Exception("puppy doggy".into())
+            GhcMessage::Exception(Arc::new("puppy doggy"))
         );
     }
 }
diff --git a/src/ghci/parse/ghc_message/mod.rs b/src/ghci/parse/ghc_message/mod.rs
index 6036a1d..34e2692 100644
--- a/src/ghci/parse/ghc_message/mod.rs
+++ b/src/ghci/parse/ghc_message/mod.rs
@@ -1,6 +1,7 @@
 //! Parser for GHC compiler output.
 
 use std::fmt::Display;
+use std::sync::Arc;
 
 use camino::Utf8PathBuf;
 use miette::miette;
@@ -75,7 +76,7 @@ pub enum GhcMessage {
     /// ```text
     /// *** Exception: /Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)
     /// ```
-    Exception(String),
+    Exception(Arc<str>),
     /// A configuration file being loaded.
     ///
     /// ```text
@@ -336,7 +337,7 @@ mod tests {
                         ].join("\n"),
                     },
                 ),
-                GhcMessage::Exception("/Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)".to_owned()),
+                GhcMessage::Exception(Arc::new("/Users/.../dist-newstyle/ghc82733_tmp_1/ghc_tmp_34657.h: withFile: does not exist (No such file or directory)")),
                 GhcMessage::Summary(
                     CompilationSummary {
                         result: CompilationResult::Err,
```

</details>

lf-:
This has been a draft for several months, so I think we can probably close it? Feel free to reopen if you'd like to finish it up though!

9999years:
@lf- Yeah I guess that's fair. I would like this feature to work, but ... until I have a way to test it...

PR #368, opened 2025-07-14
Add the ability to track warnings through re-compilation

jkeuhlen:

This PR adds support for a "Track warnings" feature which keeps track of warnings that might disappear due to GHC's recompilation processes: 
image

 GHC has a behavior that causes "ephemeral warnings" to disappear during development. When a file with warnings is recompiled due to dependency changes (without the file itself
  changing), GHC only re-emits warnings for files that were directly modified. This means warnings from unchanged files vanish from the output, making it easy for developers to
  lose track of warnings that still need to be addressed.

  Example scenario:
  1. ModuleA.hs has an unused import warning
  2. ModuleB.hs (dependency) is modified
  3. GHC recompiles both modules but only shows warnings for ModuleB.hs
  4. The warning in ModuleA.hs disappears from output despite still being present

  Solution

  This PR introduces a --track-warnings flag that maintains an in-memory store of warnings per file. The system:

  - Captures warnings during compilation and stores them by file path
  - Persists warnings across dependency-driven recompilations when files haven't changed
  - Clears warnings when files are directly modified or removed
  - Prevents duplicates by filtering out warnings already shown in current compilation
  - Integrates with error logs so tracked warnings appear in editor-compatible output files

  Usage

  # Enable warning tracking
  ghciwatch --track-warnings

  # Or via environment variable
  GHCIWATCH_TRACK_WARNINGS=1 ghciwatch

  The feature is off by default to maintain backward compatibility. When enabled, developers get a complete view of all warnings across their codebase, eliminating the
  frustrating experience of warnings disappearing during active development.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -83,6 +83,13 @@ pub struct Opts {
     #[arg(long, hide = true)]
     pub tui: bool,
 
+    /// Track warnings across recompilations.
+    ///
+    /// When enabled, warnings will be preserved in memory even when files are recompiled
+    /// due to dependency changes, helping prevent "ephemeral warnings" from being missed.
+    #[arg(long, env = "GHCIWATCH_TRACK_WARNINGS")]
+    pub track_warnings: bool,
+
     /// Generate Markdown CLI documentation.
     #[cfg(feature = "clap-markdown")]
     #[arg(long, hide = true)]
--- a/src/ghci/mod.rs
+++ b/src/ghci/mod.rs
@@ -9,6 +9,7 @@ use std::borrow::Borrow;
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::collections::BTreeSet;
+use std::collections::HashSet;
 use std::fmt::Debug;
 use std::io::IsTerminal;
 use std::process::ExitStatus;
@@ -28,6 +29,7 @@ use eyre::eyre;
 use eyre::WrapErr;
 use nix::unistd::Pid;
 use tokio::io::AsyncBufReadExt;
+use tokio::io::AsyncWriteExt;
 use tokio::io::BufReader;
 use tokio::sync::mpsc;
 use tracing::instrument;
@@ -53,6 +55,7 @@ pub mod parse;
 use parse::parse_eval_commands;
 use parse::CompilationResult;
 use parse::EvalCommand;
+use parse::ModulesLoaded;
 use parse::ShowPaths;
 
 mod ghci_command;
@@ -76,6 +79,11 @@ use file_classifier::ReloadActions;
 
 mod loaded_module;
 use loaded_module::LoadedModule;
+
+mod warning_formatter;
+
+mod warning_tracker;
+use warning_tracker::WarningTracker;
 
 use crate::aho_corasick::AhoCorasickExt;
 use crate::buffers::LINE_BUFFER_CAPACITY;
@@ -126,6 +134,8 @@ pub struct GhciOpts {
     pub stderr_writer: GhciWriter,
     /// Whether to clear the screen before reloads and restarts.
     pub clear: bool,
+    /// Whether to track warnings across recompilations.
+    pub track_warnings: bool,
 }
 
 impl GhciOpts {
@@ -204,6 +214,7 @@ impl GhciOpts {
                 stdout_writer,
                 stderr_writer,
                 clear: opts.clear,
+                track_warnings: opts.track_warnings,
             },
             tui_reader,
         ))
@@ -272,6 +283,8 @@ pub struct Ghci {
     command_handles: Vec<JoinHandle<eyre::Result<ExitStatus>>>,
     /// Monotonic counter for generating unique sync barrier nonces.
     sync_nonce: u64,
+    /// Warning tracker for managing warnings across recompilations.
+    warning_tracker: WarningTracker,
 }
 
 impl Debug for Ghci {
@@ -405,6 +418,7 @@ impl Ghci {
             },
             command_handles,
             sync_nonce: 0,
+            warning_tracker: WarningTracker::new(),
         })
     }
 
@@ -452,6 +466,13 @@ impl Ghci {
         // Get the initial list of eval commands.
         self.refresh_eval_commands().await?;
 
+        // For initialization, consider all targets as "changed"
+        self.warning_tracker.reset_changed_files();
+        for target in self.targets.iter() {
+            self.warning_tracker
+                .mark_file_changed(target.path().clone());
+        }
+
         Ok(())
     }
 
@@ -472,8 +493,15 @@ impl Ghci {
         kind_sender: oneshot::Sender<GhciReloadKind>,
     ) -> eyre::Result<()> {
         let start_instant = Instant::now();
-        let actions = self.get_reload_actions(events)?;
+        let actions = self.get_reload_actions(events.clone())?;
         let _ = kind_sender.send(actions.kind());
+
+        // Track which files were directly changed in this reload
+        self.warning_tracker.reset_changed_files();
+        for event in &events {
+            let path = self.classifier.relative_path(event.as_path())?;
+            self.warning_tracker.mark_file_changed(path);
+        }
 
         if actions.needs_restart() {
             self.opts.clear();
@@ -847,6 +875,9 @@ impl Ghci {
         }
 
         self.clear_eval_commands_for_paths(paths).await;
+        if self.opts.track_warnings {
+            self.warning_tracker.clear_warnings_for_paths(paths);
+        }
 
         Ok(())
     }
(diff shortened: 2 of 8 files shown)


src/cli.rs
```diff
@@ -20,32 +20,43 @@ use crate::normal_path::NormalPath;
 ///
 /// Load `cabal v2-repl` and watch for changes in `src`:
 ///
-///     ghciwatch
+/// ```bash
```
9999years, line 23: These examples don't use triple backticks because they're rendered in `ghciwatch --help` as well as being used for the online docs... I want to get a better system in place but don't really know what it should look like

jkeuhlen: Without this, the doctests seem to fail locally for me. Is that expected?

src/ghci/mod.rs
```diff
@@ -932,9 +993,387 @@ impl Ghci {
...
+        }
+
+        colored_lines.join("\n")
+    }
+
+    /// Apply colors to a single line of a warning message based on GHC patterns.
+    fn color_message_line(&self, line: &str, severity: parse::Severity) -> String {
+        // Detect different types of lines and apply appropriate coloring
+
+        // Source code lines with line numbers (e.g., "  28 | import Data.Coerce (coerce)")
```
9999years, line 1223: We'll probably want to factor this out into an actual parser and/or another file.

src/ghci/mod.rs
```diff
@@ -932,9 +993,387 @@ impl Ghci {
...
+                            format!("{}", flag.if_supports_color(Stdout, |text| text.red()))
+                        }
+                    };
+                    result = result.replace(flag, &colored_flag);
+                }
+            }
+        }
+
+        // // Look for [GHC-xxxx] patterns
+        // if line.contains("[GHC-") {
```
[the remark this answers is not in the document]
jkeuhlen: Yeah I tend to use "draft PR" as a signal for "this isn't ready for review yet" since I haven't even had a chance to complete my self review yet, but thanks for jumping on it so quickly!

src/cli.rs
9999years: I'm not sure there's any actual doctests? We may need some attribute to tell it to not run code examples here.

src/ghci/mod.rs
```diff
@@ -890,12 +931,36 @@ impl Ghci {
...
+            let warning_count = if self.opts.track_warnings {
+                self.warning_count()
+            } else {
+                0
+            };
+
+            if warning_count > 0 {
+                tracing::info!(
+                    "{} Finished {} in {:.2?} ({} warning{} tracked)",
+                    "Compilation succeeded".if_supports_color(Stdout, |text| text.yellow()),
```
jkeuhlen, line 949: One big design change here is that I purposefully picked a new phrase + color to signify "Compilation has finished successfully, but there are warnings present." Yellow + the succinct message here (distinct from the normal "All good!" is meant to signify that.

src/ghci/warning_formatter.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub fn colorize_message(&self, message: &str, severity: Severity) -> String {
+        let lines: Vec<&str> = message.lines().collect();
+        let mut colored_lines = Vec::new();
+
+        for line in lines {
+            let colored_line = self.colorize_line(line, severity);
+            colored_lines.push(colored_line);
+        }
+
+        colored_lines.join("\n")
```
9999years, line 35: I think you can replace most of this with a single iterator pipeline:

```
message.lines()
    .map(|line| self.colorize_line(line, severity))
    // .collect::<Vec<_>>() // Might be necessary?
    .join("\n")
```

If that doesn't work, I think you can at least avoid the first collection:

```
for line in message.lines() {
    // ... unchanged ...
}
```

src/ghci/warning_formatter.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// This method recognizes different types of lines in GHC output and applies
+    /// appropriate coloring:
+    /// - Source code lines with line numbers (e.g., "  28 | import Data.Coerce (coerce)")
+    /// - Caret indicator lines (e.g., "     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^")
+    /// - Warning/error flags in brackets (e.g., "[-Wunused-imports]")
+    fn colorize_line(&self, line: &str, severity: Severity) -> String {
+        // Detect different types of lines and apply appropriate coloring
+
+        // Source code lines with line numbers (e.g., "  28 | import Data.Coerce (coerce)")
+        if let Some(pipe_pos) = line.find(" | ") {
```
9999years, line 49: This is kinda hacky, there should at least be a comment here. Ideally we'd integrate this into the parser but I understand that's a more invasive change.

src/ghci/warning_tracker.rs
```diff
@@ -0,0 +1,335 @@
...
+        self.current_changed_files.insert(path);
+    }
+
+    /// Update warnings from a compilation log.
+    ///
+    /// This method implements smart warning persistence logic:
+    /// - Files that were directly changed: always update warnings (or clear if none)
+    /// - Files that were recompiled due to dependencies: only update if warnings exist
+    /// - Files that weren't recompiled: keep existing warnings
+    pub fn update_warnings_from_log(&mut self, log: &CompilationLog) {
```
9999years, line 52: I think this is the only method that uses `current_changed_files`, would it make sense to make that variable a parameter to this function instead of a field on the struct?

src/ghci/warning_formatter.rs
jkeuhlen: Yeah if there was a struct I could plug into with the various pieces of a GHC message, that would be really helpful. We could store original color information there as well. 

I'd like to put that off for a follow up (and left one comment-based pointer for that at the top here). Do you think another comment here would suffice for now?

src/ghci/warning_formatter.rs
jkeuhlen: Yeah that's much cleaner. I'll push a fix for it.

src/ghci/warning_tracker.rs
jkeuhlen: I could go either way? Do you feel strongly one way or the other? 

My thinking here is that the struct keeps track of the things relevant for the feature: warnings and the files currently being changed (to determine which warnings are displayed). The latter _could_ be thought of as an implementation detail but I wasn't sure if we'd want to surface the file list elsewhere externally too in the future.

src/ghci/warning_formatter.rs
jkeuhlen: Done!

src/ghci/mod.rs
```diff
@@ -932,9 +997,234 @@ impl Ghci {
         Ok(())
     }
 
+    /// Clear warnings for the given file paths.
```
9999years, line 1000: Please remove these wrapper functions and just use the methods directly.

src/ghci/mod.rs
```diff
@@ -932,9 +997,234 @@ impl Ghci {
...
+
+        for file_warnings in self.warning_tracker.get_all_warnings().values() {
+            for warning in file_warnings {
+                self.display_single_warning(warning).await;
+            }
+        }
+    }
+
+    /// Display a single warning with GHC-style formatting and colors.
+    async fn display_single_warning(&self, warning: &parse::GhcDiagnostic) {
```
9999years, line 1078: This should be a method on `GhcDiagnostic`.

src/ghci/warning_formatter.rs
```diff
@@ -0,0 +1,232 @@
...
+//! This module provides functionality to colorize GHC diagnostic messages
+//! to match the output format and colors used by GHC itself.
+
+use owo_colors::OwoColorize;
+use owo_colors::Stream::Stdout;
+
+use crate::ghci::parse::Severity;
+
+/// Formats warning and error messages with GHC-compatible colors.
+pub struct WarningFormatter;
```
9999years, line 12: This has no data, so these can just be functions. You can remove this type entirely.

src/ghci/warning_formatter.rs
jkeuhlen: Added a comment here.

src/ghci/warning_formatter.rs
```diff
@@ -0,0 +1,208 @@
...
+/// - Caret indicator lines (e.g., "     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^")
+/// - Warning/error flags in brackets (e.g., "[-Wunused-imports]")
+///
+/// TODO: Connect it to the ANSI colors
+/// that we need to preserve in <https://github.com/MercuryTechnologies/ghciwatch/blob/TrackWarnings/src/ghci/parse/ghc_message/mod.rs#L153>
+fn colorize_line(line: &str, severity: Severity) -> String {
+    // Detect different types of lines and apply appropriate coloring
+
+    // Source code lines with line numbers (e.g., "  28 | import Data.Coerce (coerce)")
+    // TODO: This is a pretty hacky way to find the lines to color. In the future, if we have all of this structured from the parser, we can store the original colors alongside the warning text and re-emit it directly.
```
9999years, line 37: Yeah, we should emit this from the parser, we're parsing it anyways

PR #369, opened 2025-08-20
Pin actions to hashes using pinact [workflow-enforcer]

mortenson:
This PR pins action references to commit hashes to mitigate supply chain attacks where a bad actor will push a new tag or override an existing tag, leading to us running malicious code immediately without explicitly updating.

We already use pinning in the backend, frontend, and infra repos.

Documentation on how to use pinact can be found here: https://developers.dmz.internal.mercury.com/docs/backend/pinning-github-action-hashes


9999years (approved):
LGTM, is the CI failure significant?

Issue #360, opened 2025-03-12
[DUX-3131] Testing UX


9999years:
The testing framework is really tedious; writing and updating tests involves pawing through pages and pages of highly verbose logging output.

<sub>From [SyncLinear.com](https://synclinear.com) | [DUX-3131](https://linear.app/mercury/issue/DUX-3131/testing-ux)</sub>

PR #393, opened 2026-03-12
tests: reduce verbose logging during test runs

kallevmercury:
For: https://github.com/MercuryTechnologies/ghciwatch/issues/360

#### Before

https://github.com/user-attachments/assets/97414dd5-6656-4a8d-9108-9aeaeb257a98

### After

https://github.com/user-attachments/assets/524b027a-cf87-400f-8263-260d4146a303


9999years (changes requested):
Info-level log messages (and above) are displayed to the user by default, so we keep as much logging as possible at debug-level or lower.

If you think that these messages should be displayed to users, I'm willing to hear you out on that, but I'd like to see screen recordings of a `ghciwatch` session running as well.

That said, the test output is _way_ too verbose, as you've found.
The approach I'd like to take here is adjusting the default log filters to be extremely minimal and narrow and then having individual tests ask for the specific log filters they need directly (I may take a shot at this myself to see how difficult it is).

https://github.com/MercuryTechnologies/ghciwatch/blob/b8bb5ae1e889d2be7415d686ced883364fdcf720/test-harness/src/ghciwatch.rs#L291-L294

I think I'm going to close this for now, let me know if you have any questions.

PS: Use `cargo nextest run` to run the tests with a faster/nicer interface.

9999years:
> The approach I'd like to take here is adjusting the default log filters to be extremely minimal and narrow and then having individual tests ask for the specific log filters they need directly (I may take a shot at this myself to see how difficult it is).

Took a shot at this, it didn't work. Almost every test expects to see ghciwatch finish reloading by looking at the span close event for the `reload` span from `ghciwatch::ghci`, but if we enable that debug-level span, we suddenly get nearly every debug-level event from ghciwatch.

The other idea I've had is having separate log filters for JSON and terminal output, so we can render events to the JSON log (for tests) which don't show up on the terminal.

9999years:
Got this done in #425 + #428, turned out to be pretty easy!

PR #396, opened 2026-03-12
CI: Replace Garnix with GHA

9999years:
Bye Garnix, enjoy being a part of, uh, Shopify? Does Shopify use Nix?

Apparently Shopify uses Nix, interesting! https://shopify.engineering/shipit-presents-how-shopify-uses-nix
https://fulltimenix.com/episodes/nix-at-shopify-ede45e30-471e-4e01-9038-70f236cf6eb8

https://garnix.io/blog/shutting-down/

image


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,39 @@
+---
+name: CI
+
+on:
+  pull_request:
```
kallevmercury, line 5: Should this also run on pushes to main?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,39 @@
...
+  ci:
+    name: nix flake check (${{ matrix.platform.runner }})
+    runs-on: ${{ matrix.platform.runner }}
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - runner: ubuntu-latest
+            system: x86_64-linux
```
kallevmercury: This might be WIP, but it seems like the `system` is not used and could be condensed to:
```yaml
platform:
  - runner: ubuntu-latest
  - runner: macos-latest
```

garnix.yaml
kallevmercury, line 1: .config/nextest.toml still references garnix, should that be updated too?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,39 @@
...
+          - runner: macos-latest
+            system: aarch64-darwin
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - uses: cachix/install-nix-action@7be5dee1421f63d07e71ce6e0a9f8a4b07c2a487 # v31.6.1
```
kallevmercury, line 29: Would a caching step after this one be beneficial?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+          extra_nix_config: |
+            extra-experimental-features = nix-command flakes
+            accept-flake-config = true
```
kallevmercury, line 33:
```suggestion
```
Isn't that setting a no-op since the nixConfig was removed from flake.nix?

9999years: Yeah, but hopefully we'll have more config (for a new cache) in the future.

.github/workflows/ci.yaml
9999years: Yes, but we don't have a public/OSS Nix cache yet.

See: https://linear.app/mercury/issue/OSPO-23/oss-nix-cache

dtpowl (approved):

parsonsmatt:
lol this PR is failing because requied checks havn't been reported yet

think we need to update the repo settings?

PR #397, opened 2026-03-14
fix(warning-tracker): ensure only warnings get stored not errors

kallevmercury:


9999years (approved):
lgtm, ty

PR #398, opened 2026-03-17
Batch file modification events to reduce compile looping

parsonsmatt:
Currently, `ghciwatch` stores events in a queue. If you have a long running compile going, this causes ghciwatch to collect file watch events, and then issue an `:add` and reload for each item in the queue. With AI agents, it is not uncommon for them to make an edit to a dozen files, each in rapid succession, *but* not so rapid that ghciwatch registers it as a single event. This queues up a dozen recompiles! Worse yet, sometimes these recompile events don't even add any new modules to the module set. For a large enough module set, we're waiting ~8-15 seconds per queue item.

This PR combines events in the queue, issuing a combined reload and `:add`. This significantly reduces the number of spurious compiles that must happen, which significantly reduces real latency in workflows that involve editing multiple files rapidly.

I've tested this locally on the mwb codebase and it works like a charm. Vastly better UX.

### DIFF
--- a/src/ghci/manager.rs
+++ b/src/ghci/manager.rs
@@ -97,6 +97,21 @@ pub async fn run_ghci(
             }
         };
 
+        // Greedily drain any additional pending events and merge them, so that rapid successive
+        // writes produce a single recompile instead of one per event.
+        loop {
+            match receiver.try_recv() {
+                Ok(pending_event) => {
+                    tracing::debug!(?pending_event, "Merging pending event");
+                    event.merge(pending_event);
+                }
+                Err(mpsc::error::TryRecvError::Empty) => break,
+                Err(mpsc::error::TryRecvError::Disconnected) => {
+                    return Err(miette!("ghci event channel closed"));
+                }
+            }
+        }
+
         // This channel notifies us what kind of reload is triggered, which we can use to inform
         // our decision to interrupt the reload or not.
         let (reload_sender, reload_receiver) = oneshot::channel();
@@ -129,6 +144,13 @@ pub async fn run_ghci(
                     // Send a SIGINT to interrupt the reload.
                     // NB: This may take a couple seconds to register.
                     ghci.lock().await.send_sigint().await?;
+                } else {
+                    // Don't interrupt, but don't lose the event either.
+                    maybe_event = Some(new_event);
+                    // Wait for the current task to finish before the next iteration.
+                    let ret = task.await;
+                    ret.into_diagnostic()??;
+                    tracing::debug!("Finished dispatching ghci event (new event queued)");
                 }
             }
             ret = &mut task => {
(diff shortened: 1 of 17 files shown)


tests/event_coalesce.rs
```diff
@@ -0,0 +1,93 @@
...
+/// follow-up compile, rather than triggering one compile per edit.
+///
+/// Scenario: a compile is running, we make 3 edits before it finishes, and
+/// expect exactly one more compile afterward containing all 3 edits.
+///
+/// Uses `--no-interrupt-reloads` so the in-progress compile runs to completion
+/// while events queue up, exercising the event preservation in the `else`
+/// branch and the greedy drain that merges pending events.
+#[test]
+async fn edits_during_compile_are_batched() {
```
parsonsmatt, line 21: I had it run these tests without the fixes, and it did fail then. The first round of tests passed. Gotta give it a feedback loop.

src/ghci/manager.rs
```diff
@@ -97,6 +97,21 @@ pub async fn run_ghci(
...
+                Ok(pending_event) => {
+                    tracing::debug!(?pending_event, "Merging pending event");
+                    event.merge(pending_event);
+                }
+                Err(mpsc::error::TryRecvError::Empty) => break,
+                Err(mpsc::error::TryRecvError::Disconnected) => {
+                    return Err(miette!("ghci event channel closed"));
+                }
+            }
+        }
```
parsonsmatt, line 113: This is the fundamental change that I'd like to land - or something like it. This prevents several queued rebuilds with minimal changed module sets, and instead we get a single rebuild with all of the changed modules during the rebuild.

src/ghci/manager.rs
```diff
@@ -129,6 +144,13 @@ pub async fn run_ghci(
                     // Send a SIGINT to interrupt the reload.
                     // NB: This may take a couple seconds to register.
                     ghci.lock().await.send_sigint().await?;
+                } else {
+                    // Don't interrupt, but don't lose the event either.
+                    maybe_event = Some(new_event);
+                    // Wait for the current task to finish before the next iteration.
+                    let ret = task.await;
+                    ret.into_diagnostic()??;
+                    tracing::debug!("Finished dispatching ghci event (new event queued)");
```
parsonsmatt, line 153: I'm less confident that this is necessary.

parsonsmatt (commented):
I've been using this in mwb today and yesterday and it's been a huge UX improvement. No more dreaded recompile queue looping!

tests/data/slow-compile/Makefile
```diff
@@ -0,0 +1,46 @@
+CABAL_OPTS ?=
```
9999years, line 1: A little annoying to have this duplicated but I'm not sure if we can do anything about it

9999years (approved):
I think this looks OK? It's basically impossible to tell if there's subtle bugs in any modifications to `src/ghci/manager.rs`, but if it performs OK locally maybe it's OK to merge

9999years:
I got this fixed recently!

PR #400, opened 2026-04-15
CARDI-731 Extract out `FileClassifier`

9999years:
When ghciwatch fails to start the underlying GHCi session (because of a bulid failure in a dependent package), it currently exits. We'd like it to wait around and attempt to restart GHCi after files change.

The logic for determining which file changes are worth restarting for is currently inside the `Ghci` struct, meaning we can't use it once the GHCi session dies.

Let's extract it so we can use it in the `Ghci` manager.

### DIFF
--- a/src/ghci/file_classifier.rs
+++ b/src/ghci/file_classifier.rs
@@ -1,0 +1,174 @@
+//! Classifies file events into reload actions based on glob patterns.
+
+use std::collections::BTreeSet;
+use std::path::Path;
+
+use camino::Utf8PathBuf;
+
+use crate::event_filter::FileEvent;
+use crate::haskell_source_file::is_haskell_source_file;
+use crate::ignore::GlobMatcher;
+use crate::normal_path::NormalPath;
+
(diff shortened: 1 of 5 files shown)


src/ghci/file_classifier.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// Classify a set of file events into reload actions.
+    ///
+    /// `targets` is the set of currently loaded modules. Pass `&ModuleSet::default()` when
+    /// GHCi has not yet initialized (all Haskell modifications will be classified as "add").
+    pub(crate) fn classify(
+        &self,
+        events: BTreeSet<FileEvent>,
+        targets: &ModuleSet,
+    ) -> miette::Result<ReloadActions> {
+        // Once we know which paths were modified and which paths were removed, we can combine
```
9999years, line 63: Cleanly moved out of `src/ghci/mod.rs`.

hdgarrood (approved):

PR #403, opened 2026-04-15
Handle compilation failures in dependencies during startup 

9999years:
When you're running ghciwatch and code in your package fails to compile, ghciwatch tells you that a reload failed and waits for files to change to try compiling again (it even does this during the initial load/startup).

However, if the compilation error is in a _dependency_ of your package, in particular one listed in the `packages` field of your `cabal.project` (i.e. with its source code in your repository), ghciwatch just says that startup failed and quits out.

Well no more!! Now ghciwatch will wait for files to change; when they do, it will attempt to start ghci again. Nice for multi-package projects!

Note that you still need to tell ghciwatch _which_ files are worth restarting over. If your package depends on a dependency `my-dependency`, you probably want to add something like this to your ghciwatch invocation:

```
ghciwatch \
    --watch my-dependency \
    --restart-glob 'my-dependency/src/**/*.hs' \
    --restart-glob 'my-dependency/my-dependency.cabal'
```

### DIFF
--- a/src/ghci/manager.rs
+++ b/src/ghci/manager.rs
@@ -1,6 +1,7 @@
 //! Subsystem for [`Ghci`] to support graceful shutdown.
 
 use std::collections::BTreeSet;
+use std::process::ExitStatus;
 use std::sync::Arc;
 
 use miette::miette;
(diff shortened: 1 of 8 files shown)


src/ghci/manager.rs
```diff
@@ -59,27 +62,76 @@ pub async fn run_ghci(
...
         .await
         .wrap_err("Failed to start `ghci`")?;
 
     // Wait for ghci to finish loading.
     let mut log = CompilationLog::default();
-    tokio::select! {
+    // Use biased select with exited_receiver before startup_result. When ghci dies, its stdout
+    // closes and read_until enters a yield loop (returning Ok(None) on each EOF read), so
+    // startup_result never completes. exited_receiver.recv() fires once GhciProcess detects the
+    // exit, and with biased polling it is guaranteed to win once a message is available.
```
hdgarrood, line 76: 👍 good comment

hdgarrood (approved):
nice, this all makes sense to me

PR #406, opened 2026-04-16
CARDI-737 Skip before-restart hooks after startup failures

9999years:
With #403, if a `--before-restart-ghci` hook was provided, we would attempt to write the hook to the `ghci` stdin handle after startup failures, which would fail because the stdin handle has been closed, causing ghciwatch to error out.


dtpowl (approved):
Nice!

PR #414, opened 2026-04-17
Use `$TIMEOUT_MULT` for tests, not retries

9999years:
This adds a new `$TIMEOUT_MULT` environment variable for tests, which acts as a numeric multiplier for all test timeouts, which can be used on CI or slower developer workstations in order to accommodate slower execution of the Haskell compilation that underlies ghciwatch integration tests.

### DIFF
--- a/nix/packages/ghciwatch.nix
+++ b/nix/packages/ghciwatch.nix
@@ -229,6 +229,9 @@
       NEXTEST_PROFILE = "ci";
       NEXTEST_HIDE_PROGRESS_BAR = "true";
 
+      # See `../../tests/README.md`.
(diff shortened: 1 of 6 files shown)


nix/packages/ghciwatch.nix
```diff
@@ -229,6 +229,9 @@
       NEXTEST_PROFILE = "ci";
       NEXTEST_HIDE_PROGRESS_BAR = "true";
 
+      # See `../../tests/README.md`.
+      TIMEOUT_MULT = "12";
```
9999years, line 233: Chosen based on fair guesstimate.

hdgarrood (approved):

PR #417, opened 2026-04-17
manager: Simplify implementation

9999years:
Behavior-preserving refactor for one of ghciwatch's nastiest modules.

Got a local review from Claude (how can we set up Claude reviews in repos like this?) and it said that the behavior was identical, so there's some weak verification.

### DIFF
--- a/src/ghci/manager.rs
+++ b/src/ghci/manager.rs
@@ -87,161 +87,31 @@ pub async fn run_ghci(
             None
         }
     };
-    if let Some(mut status) = startup_exit {
-        tracing::warn!(
-            %status,
-            "ghci exited during startup; waiting for a file change to restart"
-        );
-        loop {
-            tokio::select! {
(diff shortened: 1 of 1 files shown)


src/ghci/manager.rs
```diff
@@ -290,23 +367,55 @@ enum RetryResult {
...
+    async fn restart(&mut self) -> miette::Result<()> {
+        match self {
+            Self::Startup(ghci) => ghci
+                .startup_restart()
+                .await
+                .wrap_err("Failed to restart ghci after startup failure"),
+            Self::Runtime(ghci) => ghci
+                .lock()
+                .await
+                .restart()
```
9999years, line 395: This is a bug, it should be `.startup_restart()`, but I want this PR to preserve all behavior so that I can fix it in a separate PR with a reviewable diff.

dtpowl (approved):

PR #420, opened 2026-04-17
CARDI-745 manager: Always do `startup_restart` when `ghci` exits

9999years:
If `ghci` exits, we can't run before-restart GHCi commands, because there's no `ghci` session to send them too, so the normal `.restart()` fails.

This fixes a bug with dependency compilation errors on startup.


wavewave (approved):
Nice catch!

PR #434, opened 2026-04-27
DUX-5093 manager: fix deadlock by looping select! in handle_event

9999years:
When a non-interruptible reload was in progress and a new file event arrived, `handle_event` would return while the dispatch task still held the ghci mutex — creating a "detached" task. Subsequent dispatches would block on the mutex, and `should_interrupt` would block on the dispatch's oneshot, permanently deadlocking the main task.

Fix: wrap the `select!` in a loop. When we choose not to interrupt, accumulate the event in `pending_event` and continue waiting for the dispatch to finish. This ensures:

- No detached dispatches (`handle_event` stays until the task completes)
- No event loss (pending events are merged and retried)
- Crash/shutdown remain responsive (`select!` polls all arms each iteration)

### DIFF
--- a/src/ghci/manager.rs
+++ b/src/ghci/manager.rs
@@ -249,6 +249,12 @@ impl GhciManager {
     }
 
     /// Dispatch a watcher event, handling shutdown, interruption, and ghci death.
+    ///
+    /// Stays running until the dispatch task completes (or we decide to interrupt it),
+    /// so the spawned task never outlives this call — otherwise it could keep holding
+    /// the ghci `Mutex` and deadlock the next iteration. Events that arrive during a
+    /// non-interruptible dispatch are accumulated into `pending_event` and returned as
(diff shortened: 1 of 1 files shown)


src/ghci/manager.rs
```diff
@@ -286,30 +294,58 @@ impl GhciManager {
...
+
+                                // Cancel the in-progress reload. This releases the
+                                // `ghci` lock to prevent a deadlock.
+                                task.abort();
+
+                                // Send a SIGINT to interrupt the reload.
+                                // NB: This may take a couple seconds to register.
+                                ghci.lock().await.send_sigint().await?;
+
+                                return Ok(HandleResult::Interrupted(event));
```
9999years, line 338: These lines are moved without changes.

hdgarrood (approved):

PR #440, opened 2026-04-28
DUX-5106 miette -> eyre

9999years:
`miette::IntoDiagnostic` hides the original error from downcasting methods, which means that it's impossible (in the rare cases where we want one) to get an error _out_ of a `miette::Result` to match on it.

`IntoDiagnostic` is mandatory for error types that don't implement `miette::Diagnostic`, which (inherently) includes all of the `std` error types. I could define my own wrapper types for everything I want to match on, but that's a big hassle that just isn't worth it.

This has been an open issue for 4 years now, but we haven't really needed to match on inner errors before. Now that we do, it's a dealbreaker.

See: https://github.com/zkat/miette/issues/155

### DIFF
--- a/test-harness/Cargo.toml
+++ b/test-harness/Cargo.toml
@@ -11,8 +11,11 @@ clonable-command = "0.1.0"
 fs_extra = "1.3.0"
 futures-util = "0.3.28"
 itertools = "0.11.0"
-miette = { version = "5.9.0", features = ["fancy"] }
-nix = { version = "0.26.2", default-features = false, features = ["process", "signal"] }
+eyre = "*"
+nix = { version = "0.26.2", default-features = false, features = [
+  "process",
+  "signal",
+] }
 regex = "1.9.4"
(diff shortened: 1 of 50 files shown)


test-harness/Cargo.toml
```diff
@@ -11,8 +11,11 @@ clonable-command = "0.1.0"
 fs_extra = "1.3.0"
 futures-util = "0.3.28"
 itertools = "0.11.0"
-miette = { version = "5.9.0", features = ["fancy"] }
-nix = { version = "0.26.2", default-features = false, features = ["process", "signal"] }
+eyre = "*"
```
dtpowl, line 14: Should we pin it?

dtpowl (approved):

test-harness/Cargo.toml
9999years: It's pinned in the top-level `Cargo.toml`, so this will just resolve to the same version.

PR #444, opened 2026-04-28
DUX-5110 Don't cancel startup before writing an error log

9999years:
Previously, the `tokio::select!` in `ghci::manager` would see that the `ghci` process had exited and would cancel the `ghci.initialize()` job as a result.

However, we actually want the `ghci.initialize()` job to complete (even if it errors out), because it will read the `ghci` output and write (e.g.) compilation errors to the error log.

Usually, `ghci.initialize()` will fail with a `BrokenPipe` IO error in this case (trying to write to GHCi's stdin after the process has exited), so we catch those errors explicitly and ignore them (needs DUX-5106).

### DIFF
--- a/src/ghci/manager.rs
+++ b/src/ghci/manager.rs
@@ -68,25 +68,36 @@ pub async fn run_ghci(
         .wrap_err("Failed to start `ghci`")?;
 
     // Wait for ghci to finish loading.
+    //
+    // NB: We wait for the `ghci.initialize()` call to complete _even if_ `ghci` exits mid-way
+    // through; this lets us read the rest of its output and write a compilation log for startup
+    // errors.
     let mut log = CompilationLog::default();
-    // Use biased select with exited_receiver before startup_result. When ghci dies, its stdout
-    // closes and read_until enters a yield loop (returning Ok(None) on each EOF read), so
-    // startup_result never completes. exited_receiver.recv() fires once GhciProcess detects the
-    // exit, and with biased polling it is guaranteed to win once a message is available.
-    let startup_exit: Option<ExitStatus> = tokio::select! {
-        biased;
+    tokio::select! {
         _ = handle.on_shutdown_requested() => {
             ghci.stop().await.wrap_err("Failed to quit ghci")?;
             return Ok(());
         }
-        Some(status) = exited_receiver.recv() => Some(status),
         startup_result = ghci.initialize(&mut log, [LifecycleEvent::Startup(hooks::When::After)]) => {
             // Only reachable if ghci starts successfully (startup_result = Ok) or if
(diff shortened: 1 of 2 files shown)


src/ghci/manager.rs
```diff
@@ -68,25 +68,36 @@ pub async fn run_ghci(
...
-            None
+            match startup_result {
+                Ok(()) => {},
+                Err(err) => {
+                    let is_broken_pipe = err.chain().any(|e| {
+                        e.downcast_ref::<std::io::Error>()
+                            .is_some_and(|io| io.kind() == std::io::ErrorKind::BrokenPipe)
+                    });
+                    if is_broken_pipe {
+                        tracing::debug!("ghci stdin closed during startup (broken pipe): {err}");
```
lf-, line 92: Does this want to be a higher level than debug? Feels like the kinda thing that users maybe want to know by default ? Or maybe not?

parsonsmatt (approved):

src/ghci/manager.rs
```diff
@@ -68,25 +68,36 @@ pub async fn run_ghci(
...
+                    } else {
+                        return Err(err);
+                    }
+                }
+            }
         }
     };
+    let startup_exit: Option<ExitStatus> = exited_receiver.try_recv().ok();
```
lf-, line 100: Could simplify by matching on let Ok on the following line and deleting the `.ok()`, but shrug.

src/ghci/manager.rs
```diff
@@ -68,25 +68,36 @@ pub async fn run_ghci(
...
             // Only reachable if ghci starts successfully (startup_result = Ok) or if
             // initialization fails for a non-EOF reason (e.g., a lifecycle hook error).
-            startup_result?;
-            None
+            match startup_result {
+                Ok(()) => {},
+                Err(err) => {
+                    let is_broken_pipe = err.chain().any(|e| {
+                        e.downcast_ref::<std::io::Error>()
+                            .is_some_and(|io| io.kind() == std::io::ErrorKind::BrokenPipe)
```
lf-, line 89: do we have any way to test this? I am always scared of fucking up on these kinds of type id downcast based on my experience screwing up AnnotatedException in Haskell.

lf- (approved):
I have thought about this a bit, and I think I am convinced you *did* fix a real race condition with this, so I think it's right :)

src/ghci/manager.rs
9999years: Nah, this is expected; broken pipe = `ghci` exited, which we'll detect right after with the `exited_sender`.

src/ghci/manager.rs
9999years: yes oops https://github.com/MercuryTechnologies/ghciwatch/pull/449

